PR #1, opened 2022-01-25
Add bentheelder, hh, riaankl, and myself to OWNERS

BobyMCbobs:


OWNERS
```diff
@@ -5,3 +5,9 @@ approvers:
   - dims
   - spiffxp
   - thockin
+  - hh
+
```
ameukam, line 9: @BobyMCbobs Please, maintain the list of the approvers only to SIG Chairs and TLs.

BobyMCbobs: OK, I've updated it

PR #5, opened 2022-01-27
initial redirector service implementation

BenTheElder:
- add https://github.com/kubernetes/k8s.io/issues/1834#issuecomment-949977061 (with git "Co-authored-by" attribution), and gives it a name: αρχείο (archeío) is roughly? Greek for "registry".
- add a README with some detail
- add makefile + boilerplate scripts adapted from kubernetes-sigs/kind (we've also previously adapted these for kubernetes-sigs/kubetest2 and kubernetes/test-infra)
   - this also includes vendoring gimme from HEAD, which makes the PR look bigger than it is 🙃 
- stub in building an image with ko

I'll work more on that last part tomorrow.


cmd/archeio/main.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+}
+
+func handler(w http.ResponseWriter, r *http.Request) {
+	path := r.URL.Path
+	switch {
+	case strings.HasPrefix(path, "/v2/"):
+		doV2(w, r)
+	case strings.HasPrefix(path, "/v1/"):
+		doV1(w, r)
```
ameukam, line 44: I don't think we need to maintain compatibility for the v1. The old docker clients will be served by k8s.gcr.io.

ameukam:
Followup:
- automate image building with cloudbuild
- have presubmits running some tests
- build the infrastructure for it.

BenTheElder:
https://github.com/kubernetes-sigs/oci-proxy/issues/7 for CI
https://github.com/kubernetes-sigs/oci-proxy/pull/6 for whoops :-)

Also FYI:
With this makefile you either need `shellcheck` installed at the latest version (v0.8.0) or `docker` to run `make shellcheck` / `make verify`, the rest require just make/bash/posix and auto manage a go toolchain with `gimme` (vendored).

Other tools are managed with `hack/tools` go module, including `ko` (for distroless image builds without docker), `gotestsum` (nicer test results + junit for CI), `golangci-lint` (for linting).

The targets should otherwise be fairly self explanatory / commented, I think ...

Issue #8, opened 2022-01-27
Enable Automated Builds


BenTheElder:
1. We need to finish the `ko` build script 🙃 
2. We'll need to file a request in k8s.io for a staging registry etc.
3. Then we can enable auto-push on postsubmit with prow/cloudbuild (https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io)

BenTheElder:
1) will be easier once we know what the staging registry will be in 2)

ameukam:
We can reuse the existing GCR staging repo for k8s-infra tools. https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io/images/k8s-staging-infra-tools. 

List of the GCP accounts allowed to push: https://github.com/kubernetes/k8s.io/blob/main/groups/sig-k8s-infra/groups.yaml#L70

BenTheElder:
there's a cloudbuild for this now after https://github.com/kubernetes-sigs/oci-proxy/pull/9, but still need to do the test-infra job to automate and make sure it works there. I tried it in my own project and I think it's good to go.

BenTheElder:
https://github.com/kubernetes/test-infra/pull/25050 should have enabled this ™️ , we will see after #12 merges (or any other change)

ameukam:
First build succeeded:

```
crane digest gcr.io/k8s-staging-infra-tools/archeio:v20220128-a6b8a96
sha256:4dd742b41687c09ca9fecc7f7d18d700a6974e58d3b6ba6523a885f46c81faa7
```

PR #9, opened 2022-01-27
make image publishable with automatic date-sha tag

BenTheElder:
... using `make push-images`

will push `gcr.io/k8s-staging-infra-tools/archeio:v$date-$sha`

`make images` will build to a tarball under `bin/archeio.tar`


BenTheElder:
still need to add a cloudbuild.yaml wrapping this, and the associated prowjob https://github.com/kubernetes-sigs/oci-proxy/issues/8

Issue #7, opened 2022-01-27
Enable CI


BenTheElder:
We should enable CI to run roughly:
- `make verify`
- `make test`
- `make images build`
(each of these should probably be a distinct job in parallel)

We can probably use github actions or prow. `make verify` will require docker or the current shellcheck binary. the rest merely require bash/posix/make.

BenTheElder:
CI jobs added, need to test them, then fix any issues, then make required.

ameukam:
FWIW we already have shellcheckpresent in gcr.io/k8s-staging-infra-tools/k8s-infra:latest: https://github.com/kubernetes/k8s.io/blob/main/images/k8s-infra/Dockerfile#L38. (need to be updated)

BenTheElder:
https://github.com/kubernetes/test-infra/pull/25047#issuecomment-1023949950

these are working good now, making them blocking in https://github.com/kubernetes/test-infra/pull/25049

BenTheElder:
https://github.com/kubernetes-sigs/oci-proxy/pull/12 you can see here that this is done, we have required presubmits covering all of these things since https://github.com/kubernetes/test-infra/pull/25049

PR #11, opened 2022-01-28
fixes for enabling CI

BenTheElder:
- fix paths in verify.sh after refactoring repo layout in previous PRs
- add a script to install shellcheck in CI matching the desired version used in the shellcheck script, so we can DRY that out and just use the golang image for all presubmits
- fix minor shellcheck error that slipped through because we don't have CI enabled yet

xref #7

### DIFF
--- a/hack/make-rules/images.sh
+++ b/hack/make-rules/images.sh
@@ -40,7 +40,7 @@ cd "${REPO_ROOT}"
 
 # build images
 # TODO: bake commit info into binaries consistently
-for image in ${images[@]}; do
+for image in "${images[@]}"; do
     name="$(basename "${image}")"
     # push or local tarball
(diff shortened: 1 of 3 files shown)


BenTheElder:
(shellcheck failure is expected until this PR + https://github.com/kubernetes/test-infra/pull/25047)

BenTheElder:
(This works fine now, but I am iterating a little to minimize complexity and log spam from the install)

aojea:
> (This works fine now, but I am iterating a little to minimize complexity and log spam from the install)

:thinking: 
> tar: -: Not found in archive
> tar: Exiting with failure status due to previous errors

BenTheElder:
ok, that's better.

PR #14, opened 2022-02-04
switch to klog, simplify for pure redirect iteration

BenTheElder:
preparing for registry.k8s.io fqdn to serve *just* a redirect.
- drop unneeded logic for now
- switch logging to klog, move request logging to non-default log level (we're going to log a LOT of lines if we don't do this)

### DIFF
--- a/cmd/archeio/main.go
+++ b/cmd/archeio/main.go
@@ -17,25 +17,37 @@ limitations under the License.
 package main
 
 import (
-	"log"
+	"flag"
 	"net/http"
 	"os"
-	"regexp"
 	"strings"
+
+	"k8s.io/klog/v2"
 )
 
 func main() {
+	// cloud run expects us to listen to HTTP on $PORT
 	port := os.Getenv("PORT")
 	if port == "" {
 		port = "8080"
 	}
-	log.Printf("listening on port %s", port)
+
+	// klog setup
+	klog.InitFlags(nil)
+	flag.Parse()
+	defer klog.Flush()
+
+	// actually serve traffic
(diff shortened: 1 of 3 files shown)


cmd/archeio/main.go
```diff
@@ -17,58 +17,58 @@ limitations under the License.
...
 func main() {
+	// cloud run expects us to listen to HTTP on $PORT
 	port := os.Getenv("PORT")
 	if port == "" {
 		port = "8080"
 	}
-	log.Printf("listening on port %s", port)
+
+	// klog setup
+	klog.InitFlags(nil)
```
justinsb, line 36: Super-nit: you may want to move this to be the very first thing you do, just in we add more "above the fold" and forget that logging isn't "ready" yet

cmd/archeio/main.go
```diff
@@ -17,58 +17,58 @@ limitations under the License.
...
+	flag.Parse()
+	defer klog.Flush()
+
+	// actually serve traffic
+	klog.V(0).InfoS("listening", "port", port)
```
justinsb, line 41: Nit: any reason to use V(0) - it's the default, right?

justinsb (approved):
/lgtm

cmd/archeio/main.go
BenTheElder: I wasn't sure at first, we should probably drop the `V(0)` but it only runs once at least.

cmd/archeio/main.go
BenTheElder: filed https://github.com/kubernetes-sigs/oci-proxy/pull/15 for these

PR #19, opened 2022-03-08
Add Dependabot configuration

ameukam:
Also add GitHub Action for CodeQL.


ameukam:
cc @eddiezane @BenTheElder 

/label tide/merge-method-squash

aojea:
I'm not familiar with this tool but seems really nice, looking forward to see its results
LGTM

BenTheElder:
FWIW I disagree with updating dependencies without a known reason (as opposed to the codeQL check). I hope these PRs will be reviewed as we would in k/k.

Is there a reason we need to schedule codeQL ourselves? doesn't github already report known dependency vulnerabilities to admins privately? I see these in test-infra for example, which seems better than a public github actions job.

aojea:
> I'll remove the commit related to dependabot.
> 
> >

wow, it does updates or reports only?

agree, automating code changes is a big risk ... those should have human supervision

ameukam:
> > I'll remove the commit related to dependabot.
> > >
> 
> wow, it does updates or reports only?
> 
> agree, automating code changes is a big risk ... those should have human supervision

Dependencies update. A PR is open with the changes and the changelog. (e.g. https://github.com/kubernetes-sigs/bom/pull/64). The process is similar to what we do in k/k repo.

aojea:
> The process is similar to what we do in k/k repo.

I thought that Jordan and Dims were the bots XD

eddiezane:
I am in favor of the dependency bot since a human is approving the PR. Keeping tabs on smaller updates hopefully saves time when we need to scramble to update something for a critical update. This could also be from my Node.js background which has its own issues with dependencies...

BenTheElder:
It's worth pointing out that the only dependency we have so far is klog and there's no real reason to expect substantially more dependencies.

I'd be far more concerned about the imapacts of breakage of this project than missing the absolute latest klog.

But if the PRs are carefully reviewed, fine I guess.
IMHO it's noise unless we know we need an update. We really don't need every klog patch immediately 🙃

> CodeQL is for the archeio security analysis not the dependencies. I believe it would be interesting to that enable it knowing this repo is subject to change and grow.

I see. I am also skeptical of the value of this for a project that deals in serving unauthenticated redirects but /shrug.

I prefer tools we can run locally unless they're vulndbs. Right now we have gosec via golangci-lint which works in local development with the make targets. CI only testing is a bad developer experience IMHO and I discourage workflows that are encoded entirely in CI.

BenTheElder:
In k/k we typically discourse dependency updates without a specific need (e.g. a known bug fix) because the risk / reward is low otherwise.

BenTheElder:
Also as a Kubernetes dependency reviewer: we review the actual dependency diff more than the changelog, the changelog doesn't tell me that a dep inserted a bad init or similar. With go.mod (no vendor) + robotic changes you will have to manually go inspect this outside the PR anyhow.

ameukam:
> Also as a Kubernetes dependency reviewer: we review the actual dependency diff more than the changelog, the changelog doesn't tell me that a dep inserted a bad init or similar. With go.mod (no vendor) + robotic changes you will have to manually go inspect this outside the PR anyhow.

Let's go for manual review then.
/close

PR #21, opened 2022-03-17
Slowloris and graceful shutdown

aojea:
- protect against slowloris attacks by setting a read timeout that will close inactive connections , see https://groups.google.com/g/golang-nuts/c/MFZd6b8zQTQ/m/ISD96eyXtD4J
- implement graceful shutdown

### DIFF
--- a/cmd/archeio/main.go
+++ b/cmd/archeio/main.go
@@ -17,10 +17,13 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"flag"
 	"net/http"
 	"os"
+	"os/signal"
 	"strings"
+	"syscall"
 	"time"
 
 	"k8s.io/klog/v2"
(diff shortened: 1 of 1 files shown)


BenTheElder:
No opposition to setting a timeout, and certainly not to graceful shutdown, but I want to note:

- This app runs in cloud run behind GCLB, I don't think you can make a slowloris attack to a service behind GCLB because it won't have connected to the backend until it processes headers.
- You cannot keep a request open indefinitely because cloud run has a request timeout (default 15m)
- Go is not particularly susceptible to slowloris to begin with as noted in the golang-nuts discussion because goroutines are very cheap

ameukam:
+1 to what @BenTheElder said. We use [Cloud Armor](https://cloud.google.com/blog/products/identity-security/identifying-and-protecting-against-the-largest-ddos-attacks) for `registry.k8s.io` which has protections against to those kind of attacks.

aojea:
I'm not saying there is a problem, it is clear from your description that is not a problem, this is just another layer of security ... "be paranoid" ... when in 3 years we or someone else move the app to another place without all these protections in front we'll see it was worth :upside_down_face:

endocrimes:
It's also nice to have some defense in depth - and reqs here should be super short

BenTheElder:
Yes, I just wanted to be clear for those following along that we have thought about DoS attacks and should be in good shape.

BenTheElder (approved):
/lgtm
/approve

Issue #17, opened 2022-02-17
Monitoring of registry.k8s.io


ameukam:
`archeio` is currently deployed in sandbox mode using Cloud Run: 

We have [built-in metrics](https://cloud.google.com/run/docs/monitoring) for Cloud Run but we may need some custom metrics to gain more visibility in the system we are building. 
Possible metrics to collect:
- HTTP code for every response from k8s.gcr.io

Alerting: https://github.com/kubernetes-sigs/oci-proxy/issues/75

aojea:
Request:
- code and duration. Attributes version(v1 or v2 or unknown) and, if `path` is bounded or reasonable small (to avoid cardinality explosion) and know how many requests are used per docker image

Server:
- # active requests

aojea:
maybe path, user-agent , ... and those things are better to log in an audit.log and not as metrics, so we can parse and report on them

Top images requested, ... by user agent,

BobyMCbobs:
Added metrics in https://github.com/kubernetes-sigs/oci-proxy/pull/27

BenTheElder:
Not sure how best to export it, but cloud run collects a lot of automatic trace data, see e.g.:
<img width="1637" alt="Screen Shot 2022-03-17 at 3 59 46 PM" src="https://user-images.githubusercontent.com/917931/158907803-5431f167-979c-450c-bcde-932703467c66.png">

I'm not sure we need custom metrics.


> maybe path, user-agent , ... and those things are better to log in an audit.log and not as metrics, so we can parse and report on them

These things are already logged AFAICT:
<img width="1584" alt="Screen Shot 2022-03-17 at 4 02 49 PM" src="https://user-images.githubusercontent.com/917931/158908058-982f8e08-0f66-48e9-9de0-e03fcd809e28.png">

I don't think they're going to be particularly useful though. The volume of requests is good to know, and at some point probably the source IPs, but for now we don't want to be in the business of logging / handling PII.

As far as user agent, we already need to be OCI compliant regardless of what clients are out there. We will be as compliant as GCR.

TLDR: IMO we need to track the request volumes we are dealing with, which is something we get OOTB.

We should look at how best to open this data to the community, carefully (because potential for PII, as before with k8s.gcr.io auditing).

BenTheElder:
We have alerting in https://github.com/kubernetes/registry.k8s.io/issues/75#issuecomment-1236806204 (merging these)

[...]

PR #27, opened 2022-03-17
Add metrics

BobyMCbobs:
Add Prometheus client instrumentation for HTTP handler and new metrics port.

Metrics available:
- amount of requests
  - response code
  - method
  - version (v1, v2)
- amount of active requests

Metrics is exposed on a separate port as to not expose the info publicly.


BenTheElder:
- How is the compatibility with cloud run? I don't recall METRICS_PORT.
- Do we even need custom code to obtain metrics like # of requests with cloud run?

https://cloud.google.com/run/docs/monitoring

cmd/archeio/main.go
```diff
@@ -77,11 +138,14 @@ func handler(w http.ResponseWriter, r *http.Request) {
 	path := r.URL.Path
 	switch {
 	case strings.HasPrefix(path, "/v2/"):
+		serverMetrics.Requests.With(prometheus.Labels{"code": fmt.Sprintf("%v", http.StatusPermanentRedirect), "method": "GET", "version": "v2"}).Inc()
 		doV2(w, r)
 	case strings.HasPrefix(path, "/v1/"):
+		serverMetrics.Requests.With(prometheus.Labels{"code": fmt.Sprintf("%v", http.StatusPermanentRedirect), "method": "GET", "version": "v2"}).Inc()
```
aojea, line 144: is this always a GET?

BobyMCbobs: Whoops, that shouldn't be like that. Fixed in 3599385

BobyMCbobs:
> * How is the compatibility with cloud run? I don't recall METRICS_PORT.
> * Do we even need custom code to obtain metrics like # of requests with cloud run?
> 
> https://cloud.google.com/run/docs/monitoring

_METRICS_PORT_ is an arbitrary environment variable. I'm still figuring out the integrations

cmd/archeio/main.go
```diff
@@ -60,28 +103,50 @@ func main() {
...
 	// valid requests should be at /v2/ or /v1/, so we leave this check
 	// in the future we will selectively redirect clients to different copies
+	method := r.Method
 	path := r.URL.Path
 	switch {
 	case strings.HasPrefix(path, "/v2/"):
+		serverMetrics.Requests.With(prometheus.Labels{"code": fmt.Sprintf("%v", http.StatusPermanentRedirect), "method": method, "version": "v2"}).Inc()
 		doV2(w, r)
 	case strings.HasPrefix(path, "/v1/"):
+		serverMetrics.Requests.With(prometheus.Labels{"code": fmt.Sprintf("%v", http.StatusPermanentRedirect), "method": method, "version": "v2"}).Inc()
```
rikatz, line 145: Version=v1 in the end of the string?

BobyMCbobs: Thank you @rikatz, I missed this. Fixed in 3a50416

BobyMCbobs:
> > * How is the compatibility with cloud run? I don't recall METRICS_PORT.
> > * Do we even need custom code to obtain metrics like # of requests with cloud run?
> > 
> > https://cloud.google.com/run/docs/monitoring
> 
> _METRICS_PORT_ is an arbitrary environment variable. I'm still figuring out the integrations

@BenTheElder, I've been trying to see how to integrate custom metrics through the instrumentation in this PR into Cloud Run.
The closest thing that I've been able to find is https://cloud.google.com/monitoring/custom-metrics/creating-metrics, however this doesn't mention Cloud Run.

May I please have some clarification for on what is being asked for, in https://github.com/kubernetes-sigs/oci-proxy/issues/17?
cc @BenTheElder @ameukam

ameukam:
> May I please have some clarification for on what is being asked for, in #17? cc @BenTheElder @ameukam

@BobyMCbobs #17 was created firstly as a point of centralization of any discussion related to monitoring. This does not necessarily imply a technical action. Your PR introduces instrumentation using prometheus but the metrics introduced are already provided by the Cloud Run service like `BenTheElder`. IMHO, the PR can be closed and bring back the conversation when we are done with `registry.k8s.io` initial rollout.

BobyMCbobs:
> > May I please have some clarification for on what is being asked for, in #17? cc @BenTheElder @ameukam
> 
> @BobyMCbobs #17 was created firstly as a point of centralization of any discussion related to monitoring. This does not necessarily imply a technical action. Your PR introduces instrumentation using prometheus but the metrics introduced are already provided by the Cloud Run service like `BenTheElder`. IMHO, the PR can be closed and bring back the conversation when we are done with `registry.k8s.io` initial rollout.

@ameukam, @dims asked me (see: https://kubernetes.slack.com/archives/CCK68P2Q2/p1647550187344329?thread_ts=1647548056.298549&cid=CCK68P2Q2). But I'll circle back after the initial rollout.

/close

PR #30, opened 2022-03-25
hack: update golangci-lint

ameukam:
Also enable gosec check


ameukam:
cc @BenTheElder @dims

PR #28, opened 2022-03-21
support overriding upstream

BenTheElder:
fixes #26

Issue #24, opened 2022-03-17
full e2e testing


BenTheElder:
See also https://github.com/kubernetes-sigs/oci-proxy/issues/23

In addition to integration testing, we should start fully e2e testing a deployed copy of archeio on cloud run in an environment matching what we intend to run at registry.k8s.io.

We could do so by auto-deploying an image to registry-sandbox.k8s.io following merge of changes in this repo.

Probably this is most readily accomplished by adding it to the image building job as an additional step.

We should then make sure we have k8s e2e jobs that use registry-sandbox, and have them in a testgrid dashboard for this project.

This will give us something to look at before promoting changes to registry.k8s.io

ameukam:
/sig k8s-infra

Ref: https://github.com/kubernetes/k8s.io/issues/3411

Requirements:
  - Ensure  continuous deployment for `registry-sandbox.k8s.io`

hh:
This was mentioned as an AI in the March 30th #sig-k8s-infra meeting.

@dims requested one parallel Conformance CI job that uses registry-sandbox.k8s.io  including testgrid dashboard as we want to monitor that the sandbox redirector is production ready.

Do we have an existing job identified or do you want us to create a new one based on an existing one?

registry.k8s.io PROD is a plain redirect, and will will push hard for it to be in everything in the next release cycle.

hh:
@BenTheElder we need a staging deploy that against the staging deploy that is on a dashboard that alerts some of this group. Enough to start. (AGAINST STAGING)

Once we start merging this matching and redirecting code, staging will have the new changes and pre-rollout environment.

Possible another issue about integration testing in presubmit and spin up an integration later.

BenTheElder:
integration is done #23, we maybe have a dashboard? https://github.com/kubernetes/test-infra/pull/25964

need to get alerting, autodeploy.

BobyMCbobs:
> integration is done #23, we maybe have a dashboard? [kubernetes/test-infra#25964](https://github.com/kubernetes/test-infra/pull/25964)
> 
> need to get alerting, autodeploy.

@BenTheElder, autodeploy is currently built but waiting on #41. The permissions for the build account must have access to roles/run.admin to update the services

BenTheElder:
the dashboard is now working and caught #55, thanks all!

PR #31, opened 2022-03-28
Add unit tests

BobyMCbobs:
Adds function unit tests for archeo

### DIFF
--- a/cmd/archeio/main_test.go
+++ b/cmd/archeio/main_test.go
@@ -35,22 +35,23 @@ type suite struct {
 
 func (s *suite) runTestSuite(t *testing.T) {
 	for _, sc := range s.scenarios {
-		t.Logf("Running scenario '%v'\n", sc.name)
-		server := httptest.NewServer(s.handler)
-		defer server.Close()
-		client := server.Client()
-		if sc.request.redirect == false {
-			client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
-				return http.ErrUseLastResponse
+		t.Run(sc.name, func(t *testing.T) {
+			server := httptest.NewServer(s.handler)
+			defer server.Close()
+			client := server.Client()
+			if sc.request.redirect == false {
+				client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+					return http.ErrUseLastResponse
+				}
 			}
-		}
-		resp, err := client.Get(server.URL + sc.request.path)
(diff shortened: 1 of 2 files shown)


BobyMCbobs:
Also will be jumping to #24

ameukam:
cc @BenTheElder

cmd/archeio/main_test.go
```diff
@@ -0,0 +1,135 @@
...
+				expected: expected{url: defaultUpstreamRegistry + "/v2/", statusCode: http.StatusUnauthorized},
+			},
+			{
+				request:  request{path: "/v1/", redirect: true},
+				expected: expected{url: defaultUpstreamRegistry + "/v1/", statusCode: http.StatusNotFound},
+			},
+		},
+		tests: defaultTestFuncs(t),
+	}
+	suite.runTestSuite(t)
```
BenTheElder, line 102: can we use t.Run() for parallelism + descriptive table test names?

https://pkg.go.dev/testing#hdr-Subtests_and_Sub_benchmarks
https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

cmd/archeio/main_test.go
```diff
@@ -0,0 +1,135 @@
...
+				request:  request{path: "/v2/", redirect: false},
+				expected: expected{url: defaultUpstreamRegistry + "/v2/", statusCode: http.StatusPermanentRedirect},
+			},
+		},
+		tests: defaultTestFuncs(t),
+	}
+	suite.runTestSuite(t)
+}
+
+func TestDoV1(t *testing.T) {
```
BenTheElder, line 121: @ameukam mentioned we should drop v1 support in another PR but I didn't get that follow-up filed, do you want to do that and then drop it here? https://github.com/kubernetes-sigs/oci-proxy/pull/28#discussion_r831573854

I'll file an issue to track either way, and check back and do it if not in the morning my time 🙃

BenTheElder: filed https://github.com/kubernetes-sigs/oci-proxy/issues/34 to track

cmd/archeio/main_test.go
BobyMCbobs: Added names in f0abf43 and make parallel in 353f7ed

BobyMCbobs:
Woohoo! Thank you @dims for the merge!

Issue #23, opened 2022-03-17
integration testing


BenTheElder:
We should perform an integration test in presubmit. As any additional code complexity is added and we start to roll out registry.k8s.io to projects, it is very important that we test thoroughly, and unit testing alone will not be sufficient.

We should start by having a tool like crane pull an image through a locally running copy of archeio in an integration test target, and adding this to CI.

BenTheElder:
See also https://github.com/kubernetes-sigs/oci-proxy/issues/24 for full e2e testing post-merge.

ameukam:
Ref: https://github.com/kubernetes/k8s.io/issues/3411

BenTheElder:
done in #47

PR #32, opened 2022-03-29
Add e2e tests

BobyMCbobs:
Adds some e2e tests to target registry.k8s.io:
- Pull image `pause:3.6`
- Pull an image that doesn't exist
- Get https://registry.k8s.io/v2/ with redirects
- Get https://registry.k8s.io/v2/ without redirects

Related #24


go.mod
```diff
@@ -2,6 +2,25 @@ module sigs.k8s.io/oci-proxy
 
 go 1.17
 
-require k8s.io/klog/v2 v2.40.1
+require (
+	github.com/google/go-containerregistry v0.8.0
```
BenTheElder, line 6: let's move the test tool to it's own module, so we can clearly keep imports minimal in the main binary and not require downloading + installing test-only deps during building.

BenTheElder (commented):
I think what I'm actually looking for in #24 is mentioned by this part:

> We should then make sure we have k8s e2e jobs that use registry-sandbox, and have them in a testgrid dashboard for this project.

We should have actual kubernetes e2e tests (like kubetest + gce) running in a testgrid dashboard for this project with alerting (conformance tests maybe? or just the parallel subset?) but using registry-sandbox.k8s.io as the kubernetes image registry with the latest code from this repo deployed to registry-sandbox.k8s.io (https://github.com/kubernetes-sigs/oci-proxy/issues/24#issuecomment-1071514106 => https://github.com/kubernetes/k8s.io/issues/3411), so we know even the cloud run deployment etc is configured correctly, and that it works with a *real* client we can expect.

I think this will drive far higher confidence that we're ready to promote a change from the sandbox to production.

In #23 I think we can write something reasonable with a shell script + crane to spot check running a local registry binary and pull a known small image (pause?) through it, and add this as a make integration rule, which will give us a quicker way to test changes while developing / in presubmit, though with lower confidence. We can track crane in the hack/tools module and build and run an archeio binary locally with the make targets.

go.mod
BenTheElder: (or, perhaps rework entirely 😅 https://github.com/kubernetes-sigs/oci-proxy/pull/32#pullrequestreview-925591249)

BenTheElder:
we sort of have those e2e jobs today, but instead they're not in a dashboard we're tracking and they're using the production domain ahead of us rolling that out more widely, we're missing "auto deploy the latest code to registry-sandbox" and "run some realistic CI jobs against registry-sandbox".

I think for running against the deployed registry, we should skip straight to an actual cluster pulling from it, and only use containerregistry/crane as the client for local integration tests where it's not as reasonable to use a cluster / we're not covering the deployment configs anyhow.

PR #35, opened 2022-03-31
drop v1 support

### LINKED ISSUE #34: drop registry v1 support
This is pointless code bloat at this point apparently, GCR doesn't even serve it and it's unlikely clients were trying to use it / it doesn't support multi-arch images so the past many k8s releases probably would not really work through it ...

ref: https://github.com/kubernetes-sigs/oci-proxy/pull/28#discussion_r831571816

should be a clean, simple PR. we can do further refactors in the future.

/help
/good-first-issue

BenTheElder:
fixes #34


ameukam:
/lgtm
/approve
/hold

Remove `/hold` when ready

PR #36, opened 2022-03-31
upgrade to go1.18

BenTheElder:
we will probably want to use golang 1.18 features like net/netip package


go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/oci-proxy
 
-go 1.17
+go 1.18
```
BenTheElder, line 3: most go upgrades should *not* involve editing this, but we're actually potentially interested in 1.18+ specific functionality as a relatively new project that could benefit from it ad we don't have to worry about users building this with older go.

ameukam:
/lgtm
/approve
/hold

Remove `/hold` when ready

PR #40, opened 2022-04-01
Add auto-deploy to CloudRun regions

BobyMCbobs:
Related: https://github.com/kubernetes/k8s.io/issues/3577

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,15 +1,19 @@
 # See https://cloud.google.com/cloud-build/docs/build-config
 steps:
-  - name: gcr.io/cloud-builders/docker
+  - id: build-image
+    name: gcr.io/cloud-builders/docker
     entrypoint: "/usr/bin/make"
     args:
       - "push-images"
       - "TAG=$_GIT_TAG"
-  - name: "gcr.io/google.com/cloudsdktool/cloud-sdk"
+  - id: deploy-staging
+    name: "gcr.io/google.com/cloudsdktool/cloud-sdk"
     entrypoint: "/usr/bin/make"
     args:
       - "deploy"
       - "TAG=$_GIT_TAG"
+    waitFor:
+      - build-image
 substitutions:
   # variables set by kubernetes/test-infra/images/builder
   # set by image-builder to vYYYYMMDD-hash
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -6,27 +6,10 @@ steps:
       - "push-images"
       - "TAG=$_GIT_TAG"
   - name: "gcr.io/google.com/cloudsdktool/cloud-sdk"
-    entrypoint: "/usr/bin/gcloud"
+    entrypoint: "/usr/bin/make"
     args:
-      - "run"
-      - "services"
-      - "update"
-      - "k8s-infra-oci-proxy-us-central1"
-      - "--image"
-      - "gcr.io/k8s-staging-infra-tools/archeio:$_GIT_TAG"
-      - "--region"
-      - "us-central1"
-  - name: "gcr.io/google.com/cloudsdktool/cloud-sdk"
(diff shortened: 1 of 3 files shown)


BobyMCbobs:
What might need to be done
- cloudrun admin/update permissions for the k8s-infra build service account
- postsubmit job?

dims:
/approve

leaving LGTM for @ameukam 

@BobyMCbobs tested locally right?

BobyMCbobs:
> /approve

Thank you @dims 

> leaving LGTM for @ameukam
> 
> @BobyMCbobs tested locally right?

I ran this against k8s-infra-ii-sandbox with Run Services brought up and with updated permissions for the build GCP service account

dims:
thanks for confirming @BobyMCbobs

hack/make-rules/deploy.sh
```diff
@@ -0,0 +1,38 @@
...
+IMAGE_REPO="${IMAGE_REPO:-gcr.io/k8s-staging-infra-tools/archeio}"
+PROJECT="${PROJECT:-k8s-infra-oci-proxy}"
+
+REGIONS=(
+    us-central1
+    us-west1
+)
+
+# shellcheck disable=SC2048
+for REGION in ${REGIONS[*]}; do
```
ameukam, line 33: Why not ? 
```suggestion
for REGION in "${REGIONS[*]}"; do
```

hack/make-rules/deploy.sh
```diff
@@ -0,0 +1,38 @@
...
+    gcloud --project="${PROJECT}" \
```
ameukam, line 34: Can be done later:
```suggestion
    gcloud beta --project="${PROJECT}" \
```

We can get access to some features usable and test for the sandbox environment.

cloudbuild.yaml
```diff
@@ -1,10 +1,15 @@
...
+  - name: gcr.io/cloud-builders/docker
+    entrypoint: "/usr/bin/make"
+    args:
+      - "push-images"
+      - "TAG=$_GIT_TAG"
+  - name: "gcr.io/google.com/cloudsdktool/cloud-sdk"
+    entrypoint: "/usr/bin/make"
+    args:
+      - "deploy"
+      - "TAG=$_GIT_TAG"
```
ameukam, line 12: We could introduce an order of steps to make sure the images are built before deployment. (Also we may have separate cloudbuild files). 
Ben's feedback can be really useful.

hack/make-rules/deploy.sh
BobyMCbobs: This won't iterate over all the values in $REGIONS, but will treat all values as a single string

hack/make-rules/deploy.sh
BobyMCbobs: Which features will this enable that we need?

cloudbuild.yaml
BobyMCbobs: I'm not sure what needs to be changed, currently it does build and push the image before deploying it to the staging environments

hack/make-rules/deploy.sh
ameukam: No features needed at the moment but might be useful in the features. Feel free to ignore my comment.

cloudbuild.yaml
ameukam: Something like that https://cloud.google.com/build/docs/configuring-builds/configure-build-step-order#examples. 

Feel free to ignore my comment if it's not needed.

BobyMCbobs: My experience of using CloudBuild is that the jobs run consecutive, though it'll be pretty trivial to add this to ensure that always run in order.

BobyMCbobs: Added in 1012a65

BobyMCbobs:
Woohoo! Thank you @dims merging!

Issue #39, opened 2022-03-31
oci-proxy should redirect per-AWS region


hh:
Here are the regions (and ASNs) we need to oci-proxy to actively redirect.

@ameukam is creating the AWS infra in https://github.com/kubernetes/k8s.io/issues/3568.

If we treat Amazon ASNs not in ip-ranges.json as a type of region, then it’s #4 at 11.12% of Amazons total:

- 21.5% : us-west-2
- 16.7% : us-west-1
-  13% : us-east-1
- 11.1% : Other Amazon ASNs (not in ip-ranges.json, but collected in [k8s.io/meta/asns/amazon.yaml](https://github.com/kubernetes/k8s.io/blob/main/registry.k8s.io/infra/meta/asns/amazon.yaml)
- 11.1% : eu-central-1
- 9.2% :  eu-central-1
- 6.39% : us-east-2
- 5.32% : ap-southeast-1
- 2.99% : us-west-1
- 2.6% : ap-northeast-1
- 2.12% : ap-south-1
=====^^Roughly ~80%^^=====

![image](https://user-images.githubusercontent.com/61125752/161145295-51614363-7b67-4335-b3e4-966d2c5ed12f.png)

ameukam:
>11.1% : eu-central-1
> 9.2% : eu-central-1

Why 2 percentages for the same region ?

Riaankl:
![image](https://user-images.githubusercontent.com/61125752/161153529-e16ad1ae-7b56-44cd-98c9-690f849f13df.png)

Riaankl:
> > 11.1% : eu-central-1
> > 9.2% : eu-central-1
> 
> > 2.99% : us-west-1
> > 16.7% : us-west-1
> 
> Why 2 percentages for the same region ?

@hh typo. 
We got the graph now ^^

BenTheElder:
how did we go from 6.5% to 20%? https://github.com/kubernetes-sigs/oci-proxy/issues/20#issuecomment-1077629290

hh:
6.5% was the total across our entire spend... this number is only across our Amazon spend.

Riaankl:
80/20 Work out exactly.
6 of the 27 regions account for 80% of the traffic, and 6/27=21.4%
20% of the regions account for 80% of the traffic.
[Vilfredo Pareto](https://en.wikipedia.org/wiki/Pareto_principle) would be so proud that it still work that way!

aojea:
lol

[...]

PR #42, opened 2022-04-13
implement CIDR matching

BenTheElder:
This is inspired by how the kernel does routing (with a LPC Trie), thanks @aojea for the pointer to https://vincent.bernat.ch/en/blog/2017-ipv4-route-lookup-linux

There is one existing go library for this at https://github.com/yl2chen/cidranger but it implements more of a set type (we need a map to AWS regions), doesn't use the newer `net/netip` types which are more efficient, etc. 

We implement a reasonably efficient generic map type in relatively little code here (~260 lines TrieMap including comments, whitespace, license header etc.).

While the current implementation here lacks level and path compression, cidranger [only implements path compression](https://github.com/yl2chen/cidranger/blob/d1cb2c52f37ac257db665ddfaf4af22e89863471/trie.go#L35). 
For AWS data memory usage is acceptable without compression (ballpark less than 1MB), while lookup performance is with the AWS IPv4 data is ~40ns / op for this implementation vs ~400ns / op for cidranger on my 16" intel macbook laptop in microbenchmarking (and both _drastically_ better than brute force looping over CIDRs).

See: https://github.com/kubernetes-sigs/oci-proxy/pull/42#issuecomment-1098586853 with the full dataset ipv4 + ipv6 matching, performance is ~fine. We can continue optimizing in follow-ups as needed.

See: https://github.com/kubernetes-sigs/oci-proxy/issues/39

/cc @dims @ameukam


pkg/net/cidrs/bruteforce.go
```diff
@@ -0,0 +1,44 @@
...
+	return &bruteForceMapper[V]{
+		mapping: mapping,
+	}
+}
+
+type bruteForceMapper[V comparable] struct {
+	mapping map[netip.Prefix]V
+}
+
+func (b *bruteForceMapper[V]) GetIP(addr netip.Addr) (value V, matched bool) {
```
BenTheElder, line 37: note: named return values are the easiest way to return a default generic type cleanly while also documenting the return values

otherwise we would have a local type like `var value V` to achieve the same thing, you cannot do say `return nil, false`

BenTheElder:
preliminary benchmarking has brute force taking 150x time, even with the current relatively simplistic implementation.

it does take longer than i'd like to initialize the trie, we may want to pre-initialize it (expected, but didn't want to do that before measuring...)

BenTheElder:
NOTE: most of the lines in the diff are just `pkg/net/cidrs/aws/internal/ranges2go/zz_generated_rawdata.go` which is https://ip-ranges.amazonaws.com/ip-ranges.json stuffed into a go string by a small script. The actual diff is not that large.

BenTheElder:
```
go test -benchmem -run=^$ -coverprofile=$elided -bench . sigs.k8s.io/oci-proxy/pkg/net/cidrs/aws

goos: darwin
goarch: amd64
pkg: sigs.k8s.io/oci-proxy/pkg/net/cidrs/aws
cpu: Intel(R) Core(TM) i9-9980HK CPU @ 2.40GHz
BenchmarkNewAWSRegionMapper-16       	     786	   1487210 ns/op	  870320 B/op	   29875 allocs/op
BenchmarkNewAWSegionBruteForce-16    	   96062	     12387 ns/op	       8 B/op	       1 allocs/op
BenchmarkAWSRegionTrieMap-16         	10781480	       111.2 ns/op	       0 B/op	       0 allocs/op
BenchmarkAWSRegionBruteForce-16      	   76730	     14975 ns/op	       0 B/op	       0 allocs/op
PASS
coverage: 100.0% of statements
ok  	sigs.k8s.io/oci-proxy/pkg/net/cidrs/aws	5.771s
```

With the real AWS data loaded including IPv6 (unless I've somehow misread these results, I don't find myself benchmarking go all that often...)

~ .0015s to initialize the TrieMap and ~ 0.87Mb memory overhead (`BenchmarkNewAWSRegionMapper-16`)
~ 111ns/op lookup times.

We have room to improve these if need be, but they should be plenty acceptable already.

BenTheElder:
Also a conscious decision: we could fairly easily minify `pkg/net/cidrs/aws/internal/ranges2go/zz_generated_rawdata.go` but I chose *not* to do so, because when the data does change occasionally it will be easy to read the diff (we will see some new ranges inserted in some regions), versus it will not be so easy minified.

The file is not included at runtime anyhow, only in the repo. The raw data from AWS is pre-processed by ranges2go into pre-parsed CIDRs in a more useful format for our purposes (`map[string][]netip.Prefix` region : CIDRS).

pkg/net/cidrs/triemap.go
```diff
@@ -0,0 +1,262 @@
...
+func (t *trieMap) insertIPV4(cidr netip.Prefix, key int) {
+	// ensure root node
+	if t.ipv4Root == nil {
+		t.ipv4Root = &trieNode{}
+	}
+
+	// walk bits high to low, inserting matching ip path up to mask bits
+	curr := t.ipv4Root
+	ip := cidr.Addr().As4()
+	ipInt := uint32(ip[3]) | uint32(ip[2])<<8 | uint32(ip[1])<<16 | uint32(ip[0])<<24
```
dims, line 132: constructing the ipInt from the bytes could use a comment.

pkg/net/cidrs/triemap.go
```diff
@@ -0,0 +1,262 @@
...
+	if t.ipv6Root == nil {
+		t.ipv6Root = &trieNode{}
+	}
+
+	// walk bits high to low, inserting matching ip path up to mask bits
+	curr := t.ipv6Root
+	ip := cidr.Addr().As16()
+	bits := cidr.Bits()
+	for i := 127; i >= (128 - bits); i-- {
+		if (uint8(ip[15-i/8])>>(i%8))&1 == 1 {
```
dims, line 164: a comment here too please

pkg/net/cidrs/triemap.go
```diff
@@ -0,0 +1,262 @@
...
+	curr := t.ipv4Root
+	if curr == nil {
+		return -1, false
+	}
+	if curr.value != nil && curr.value.cidr.Contains(addr) {
+		return curr.value.key, true
+	}
+	// walk IP bits high to low, checking if current node matches
+	ip := addr.As4()
+	ipInt := uint32(ip[3]) | uint32(ip[2])<<8 | uint32(ip[1])<<16 | uint32(ip[0])<<24
```
dims, line 200: same

pkg/net/cidrs/triemap.go
```diff
@@ -0,0 +1,262 @@
...
+		return -1, false
+	}
+	if curr.value != nil && curr.value.cidr.Contains(addr) {
+		return curr.value.key, true
+	}
+	// walk IP bits high to low, checking if current node matches
+	ip := addr.As16()
+	for i := 127; i >= 0; i-- {
+		// walk based on current address bit
+		if (uint8(ip[15-i/8])>>(i%8))&1 == 1 {
```
dims, line 240: same

pkg/net/cidrs/triemap.go
dims: in the ipv6 case (lines 163/164) we are traversing the 16 bytes but here seem to create an integer out of the 4 bytes and THEN traverse it... do we want to avoid that with code similar to the ipv6 case? (curious)

pkg/net/cidrs/triemap.go
dims: same question as above (regarding having the need to construct an int)

pkg/net/cidrs/triemap.go
dims: thanks Ben!

BenTheElder: I'm splitting the IPv4 and IPv6 benchmarks so we can more clearly measure these changes.

After splitting, doing IPv4 the byte by byte way is ~ 55 ns/op vs ~ 39 ns/op pre-casting to uint32 (some variance in bench runs but consistently faster pre-casting).

IPv6 is >200 ns/op vs ~170 ns / op pre-casting to two uint64 (again variance, but faster pre-casting).

All of these values are more than excellent for our purposes, but IMHO it's more idiomatic for this type of algorithm to mask fixed width integers than byte by byte as well.

pkg/net/cidrs/triemap.go
```diff
@@ -197,10 +213,12 @@ func (t *trieMap) getIPv4(addr netip.Addr) (int, bool) {
 	}
 	// walk IP bits high to low, checking if current node matches
 	ip := addr.As4()
+	// first cast to uint32 for fast bit access
+	// NOTE: IP addresses are big endian, so the low bits are in the last byte
 	ipInt := uint32(ip[3]) | uint32(ip[2])<<8 | uint32(ip[1])<<16 | uint32(ip[0])<<24
 	for i := 31; i >= 0; i-- {
 		// walk based on current address bit
-		if (ipInt>>i)&1 == 1 {
+		if (ipInt & (uint32(1) << i)) != 0 {
```
BenTheElder, line 221: (comparing to zero is a fast built-in operation on all? CPU, the compiler can probably? optimize these predictable bitshifts, and we have to do one mask either way; benchmarks seem to agree)

BenTheElder:
I think the `unused` lint flake:
```
pkg/net/cidrs/triemap.go:53:2: field `keyToValue` is unused (unused)
	keyToValue map[int]V
	^
pkg/net/cidrs/bruteforce.go:34:2: field `mapping` is unused (unused)
	mapping map[V][]netip.Prefix
	^
pkg/net/cidrs/triemap.go:54:2: field `valueToKey` is unused (unused)
	valueToKey map[V]int
```
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_oci-proxy/42/pull-oci-proxy-verify/1514800357945905152

does not fail locally, and by trivial inspection these errors are not correct
/retest

dims:
/approve
/lgtm

thanks @BenTheElder looking forward to hooking up this to the main code base next!

PR #47, opened 2022-04-15
redirect AWS client blob requests to s3 based on client IP -> AWS region

### LINKED ISSUE #39: oci-proxy should redirect per-AWS region

BenTheElder:
builds on #42, depends on #46
~fixes #39 

Implements:
- given an HTTP request, get the client IP (either locally in development, or in cloud run from GCLB)
- map IP request to AWS region
- if not an AWS client, redirect to primary registry
- if an AWS client, redirect to a bucket copy based on the region 

Also clarifies that cmd/archeio is *not* engineered to be reusable at this time, instead the reusable bits are implemented as library packages (like the CIDR matching, IP to AWS region matching). 
We can greatly simplify runtime concerns and reliability by keeping it this way for now. 
In the future it's possible we'll be using an entirely different approach (CDN?) anyhow.

Needs cleanup, tests, and a feature gate of some sort (disabled by default, we're not ready to roll this out yet), but it's there.

Right now all regions are mapped to the single bucket we have.
We can use that to test in CI in the sandbox, but before enabling in production we will need to replace with our real bucket mapping.

### DIFF
--- a/cmd/archeio/app/buckets.go
+++ b/cmd/archeio/app/buckets.go
@@ -18,7 +18,6 @@ package app
 
 func regionToBucket(region string) string {
 	// for now always return @ameukam's test bucket
-	// TODO: add a test that we exhaustively cover all regions
 	switch region {
 	default:
 		return "https://painfully-really-suddenly-many-raccoon-image-layers.s3.us-west-2.amazonaws.com"
--- a/cmd/archeio/app/buckets.go
+++ b/cmd/archeio/app/buckets.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package app
 
 func regionToBucket(region string) string {
 	// for now always return @ameukam's test bucket
--- a/cmd/archeio/main_test.go
+++ b/cmd/archeio/main_test.go
@@ -1,0 +1,116 @@
+//go:build !nointegration
+// +build !nointegration
+
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"net/http"
+	"os"
+	"os/exec"
+	"testing"
+	"time"
+
+	"sigs.k8s.io/oci-proxy/internal/integration"
+)
+
+func TestIntegrationMain(t *testing.T) {
+	integration.MaybeSkip(t)
+
+	// setup crane
+	rootDir, err := integration.ModuleRootDir()
+	if err != nil {
+		t.Fatalf("Failed to detect module root dir: %v", err)
+	}
+	// NOTE: also ensures rootDir/bin is in front of $PATH
+	if err := integration.EnsureCrane(rootDir); err != nil {
+		t.Fatalf("Failed to ensure crane: %v", err)
+	}
+
+	// build binary
+	buildCmd := exec.Command("make", "archeio")
+	buildCmd.Dir = rootDir
+	if err := buildCmd.Run(); err != nil {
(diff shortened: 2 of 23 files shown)


BenTheElder:
moved some of this to https://github.com/kubernetes-sigs/oci-proxy/pull/48

BenTheElder:
tests still need work, will get back to this in the morning.

cmd/archeio/main_test.go
```diff
@@ -17,121 +20,99 @@ limitations under the License.
...
+		t.Fatalf("Failed to build archeio for integration testing: %v", err)
+	}
 
-type suite struct {
-	handler   http.Handler
-	scenarios []scenario
-	tests     []func(resp *http.Response, scenario scenario)
-}
+	// start server in background
+	testPort := "61337"
```
aojea, line 56: why not pick a random one with :0?

BenTheElder: :effort:
If I pick it by binding a port in the test I need to make sure to bind to the same addr and hope the port doesn't get used between releasing it and starting the server anyhow.
If I let the server process pick it then I need to discover it.

aojea: it is highly unlikely that happens, or we should have the CI failing constantly , ... not a big deal though

BenTheElder: CI doesn't have peers opening high ports because we run tests in a network namespace for those tests. Local dev may.

BenTheElder:
I'll have some more unit tests / cleanup committed later this morning and then this should be ready for reviews.

cmd/archeio/main_test.go
aojea: you can't run the test continously, `-count 2` the port closed has a default TIME_WAIT status , 1min IIRC , it should not be able to bind it

BenTheElder: > you can't run the test continously, -count 2 the port closed has a default TIME_WAIT status , 1min IIRC , it should not be able to bind it

That's not something you should be doing anyhow, this is a full-on integration test, we have unit tests in `cmd/archeio/app` testing the http handlers etc directly you can run continuously, the test here is just to exercise the binary. it's more like an e2e.

BenTheElder:
Probably should feature-gate this in some fashion, but it's otherwise done.

BenTheElder:
I think we can do that in a follow-up, this is pretty large as-is.

cmd/archeio/app/buckets.go
```diff
@@ -0,0 +1,26 @@
...
+*/
+
+package app
+
+func regionToBucket(region string) string {
+	// for now always return @ameukam's test bucket
+	// TODO: add a test that we exhaustively cover all regions
+	switch region {
+	default:
+		return "https://painfully-really-suddenly-many-raccoon-image-layers.s3.us-west-2.amazonaws.com"
```
dims, line 24: hehehehehe!!

cmd/archeio/app/buckets.go
```diff
@@ -0,0 +1,26 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package app
+
+func regionToBucket(region string) string {
+	// for now always return @ameukam's test bucket
+	// TODO: add a test that we exhaustively cover all regions
```
BenTheElder, line 21: oops, this TODO is TODOne

cmd/archeio/main_test.go
dims: +1 to merge and iterate

cmd/archeio/app/buckets.go
dims: thanks!

PR #48, opened 2022-04-19
add some more details to README.md

BenTheElder:
clarifying some things, e.g. a lot of folks seem to be missing the "join the mailinglist" part and my inbox misses most of those. so let's add some details that aren't tucked away in the doc as well.


BenTheElder:
moves over some writing from #47

PR #49, opened 2022-04-20
Redirect to a placeholder URL about this service

dims:
For now redirect to a place holder URL (github repo for this service)

Signed-off-by: Davanum Srinivas <davanum@gmail.com>

### DIFF
--- a/cmd/archeio/app/handlers.go
+++ b/cmd/archeio/app/handlers.go
@@ -26,6 +26,11 @@ import (
 	"sigs.k8s.io/oci-proxy/pkg/net/cidrs/aws"
 )
 
+const (
+	// hard coded for now
+	infoURL = "https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io"
+)
+
 // MakeHandler returns the root archeio HTTP handler
 //
 // upstream registry should be the url to the primary registry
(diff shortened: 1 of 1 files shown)


cmd/archeio/app/handlers.go
```diff
@@ -39,6 +44,8 @@ func MakeHandler(upstreamRegistry string) http.Handler {
 		switch {
 		case strings.HasPrefix(path, "/v2/"):
 			doV2(w, r)
+		case path == "/":
```
BenTheElder, line 47: we should get a unit test case for this

dims: https://github.com/kubernetes-sigs/oci-proxy/pull/51/files

cmd/archeio/app/handlers.go
```diff
@@ -26,6 +26,11 @@ import (
 	"sigs.k8s.io/oci-proxy/pkg/net/cidrs/aws"
 )
 
+const (
+	// hard coded for now
+	infoURL = "https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io"
```
BenTheElder, line 31: I think we should probably switch to https://github.com/kubernetes/k8s.io/wiki/New-Registry-url-for-Kubernetes-(registry.k8s.io)

this link doesn't seem to have much relevant, we're not using the things that are there

BenTheElder: will try to file a follow up soon. this just came up after the community update today

PR #50, opened 2022-04-20
increase pkg/net/cidrs unit test coverage to 100%

BenTheElder:
just to be on the safe side. coverage was good before at 94% for triemap.go and 100% for the rest.

### DIFF
--- a/pkg/net/cidrs/triemap.go
+++ b/pkg/net/cidrs/triemap.go
@@ -223,14 +223,14 @@ func (t *trieMap) getIPv4(addr netip.Addr) (int, bool) {
 				curr = curr.child1
 			} else {
 				// dead end
-				return -1, false
+				break
 			}
 		} else {
 			if curr.child0 != nil {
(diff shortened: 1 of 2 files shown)


pkg/net/cidrs/triemap.go
```diff
@@ -272,21 +271,20 @@ func (t *trieMap) getIPv6(addr netip.Addr) (int, bool) {
...
+				break
 			}
 		}
 		// check for a match in the current node
 		if curr.value != nil && curr.value.cidr.Contains(addr) {
 			return curr.value.key, true
 		}
 	}
-	// we walked to the final IP bit without a match
 	return -1, false
```
BenTheElder, line 289: these were unreachable / dead code before, consolidated with the dead-end returns => `break` which we always hit. I would hope it compiles the same ...

hh: 👍

hh (approved):

PR #52, opened 2022-04-20
test internal ranges2go tool

BenTheElder:
- add some unit tests
- ensure reproducible generated output (uncovered by testing, we iterated a map, get the keys and sort before iterating)
- fix the wrapper script to handle the fact that you can't `go run *.go` when you have test files
- regenerate
- add verify check that checked in file matches current generator


BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_oci-proxy/52/pull-oci-proxy-test/1516922599341297664

This brings us up to 82% unit test coverage overall.

We also have the integration test which (not measured) covers the rest of `cmd/archeio`, and we ensure the latest data is generated, so most (but not yet all...) of the "not covered" lines are exercised in non-unit testing.

PR #53, opened 2022-04-21
drive up test coverage, drop redundant integration test skip

BenTheElder:


cmd/archeio/main_test.go
```diff
@@ -33,8 +33,6 @@ import (
 // TestIntegrationMain tests the entire, built binary with an integration
 // test, pulling images with crane
 func TestIntegrationMain(t *testing.T) {
-	integration.MaybeSkip(t)
```
BenTheElder, line 36: this was a hold-over from KIND where we iterated on this previously. the test tags and regex matching should be sufficient

pkg/net/cidrs/aws/internal/ranges2go/main.go
```diff
@@ -27,11 +27,7 @@ func main() {
 		outputPath = "./zz_generated_range_data.go"
 	}
 	// parse raw AWS IP range data
-	parsed, err := parseIPRangesJSON([]byte(ipRangesRaw))
-	if err != nil {
-		panic(err)
-	}
-	rtp, err := regionsToPrefixesFromData(parsed)
+	rtp, err := regionsToPrefixesFromRaw(ipRangesRaw)
```
BenTheElder, line 30: we want to minimize the amount of logic in any main.go, and ensure we're testing everything else thoroughly with quick and cheap unit tests ...

pkg/net/cidrs/aws/internal/ranges2go/parse_test.go
```diff
@@ -77,3 +77,56 @@ func TestParseIPRangesJSON(t *testing.T) {
 		t.Fatal("expected error parsing garbage data but got none")
 	}
 }
+
+func TestRegionsToPrefixesFromData(t *testing.T) {
+	t.Run("bad IPv4 prefixes", func(t *testing.T) {
```
aojea, line 82: why not test valid prefixes too?  ... and use a table test, only expectError is tested

BenTheElder: there's other tests that cover this method, but we do not cover the error cases in those tests.

BenTheElder: table test seems excessive for N=2.

pkg/net/cidrs/aws/internal/ranges2go/main.go
dims: ack

pkg/net/cidrs/aws/internal/ranges2go/parse_test.go
dims: ack

cmd/archeio/main_test.go
dims: ack

PR #56, opened 2022-04-21
fix bugs in GetClientIP and MakeHandlers

### LINKED ISSUE #55: Docker client fails to download from the sandbox
After a auto-deployment of `archeio`, `e2e-kops-grid-gcr-mirror-canary` is failing:

See https://testgrid.k8s.io/sig-k8s-infra-oci-proxy#kops-grid-gcr-mirror-canary

```shell
❯ docker pull registry-sandbox.k8s.io/pause:3.1
3.1: Pulling from pause
67ddbfb20a22: Pulling fs layer 
error pulling image configuration: error parsing HTTP 400 response body: invalid character 'P' looking for beginning of value: "ParseAddr(\" 34.110.128.221\"): unexpected character (at \" 34.110.128.221\")\n"
```

Seems like an issue with IP address parsing

/kind bug
/sig k8s-infra
/area artifacts
/milestone v1.24

BenTheElder:
fixes #55

### DIFF
--- a/cmd/archeio/app/handlers.go
+++ b/cmd/archeio/app/handlers.go
@@ -75,6 +75,7 @@ func makeV2Handler(upstreamRegistry string) func(w http.ResponseWriter, r *http.
 		clientIP, err := getClientIP(r)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
 		}
 
 		region, matched := regionMapper.GetIP(clientIP)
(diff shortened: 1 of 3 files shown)


cmd/archeio/app/handlers.go
```diff
@@ -75,6 +75,7 @@ func makeV2Handler(upstreamRegistry string) func(w http.ResponseWriter, r *http.
 		clientIP, err := getClientIP(r)
 		if err != nil {
 			http.Error(w, err.Error(), http.StatusBadRequest)
+			return
```
BenTheElder, line 78: TODO to revisit on detecting this class of bug in tests.class of bug in tests.

BenTheElder:
just noticed we added a github action for unit tests and it doesn't work: https://github.com/BenTheElder/oci-proxy/runs/6117603602?check_suite_focus=true

will file another PR to remove this, `pull-oci-proxy-test` runs unit + integration, producing a coverage map.

PR #54, opened 2022-04-21
Remove shellcheck comment

ameukam:
Signed-off-by: Arnaud Meukam <ameukam@gmail.com>

PR #59, opened 2022-04-21
Revert "Remove shellcheck comment"

BobyMCbobs:
Reverts kubernetes-sigs/oci-proxy#54.

Fixes splitting of `$REGIONS` into `us-central1`, `us-west1` instead of `us-central1 us-west1`

### DIFF
--- a/hack/make-rules/deploy.sh
+++ b/hack/make-rules/deploy.sh
@@ -29,7 +29,8 @@ REGIONS=(
     us-west1
 )
 
-for REGION in "${REGIONS[@]}"; do
+# shellcheck disable=SC2048
+for REGION in ${REGIONS[*]}; do
     gcloud --project="${PROJECT}" \
         run services update "${SERVICE_BASENAME}-${REGION}" \
         --image "${IMAGE_REPO}:${TAG}" \


BenTheElder:
this isn't correct, the previous PR is correct:

```bash
$ # be super clear that we're in bash
$ bash
# initialize regions array
$ REGIONS=(
    us-central1
    us-west1
)
# loop regions array, same as https://github.com/kubernetes-sigs/oci-proxy/pull/54/files
$ for REGION in "${REGIONS[@]}"; do
> echo "$REGION"
> done
us-central1
us-west1
```

BobyMCbobs:
> this isn't correct, the previous PR is correct:
> 
> ```shell
> $ # be super clear that we're in bash
> $ bash
> # initialize regions array
> $ REGIONS=(
>     us-central1
>     us-west1
> )
> # loop regions array, same as https://github.com/kubernetes-sigs/oci-proxy/pull/54/files
> $ for REGION in "${REGIONS[@]}"; do
> > echo "$REGION"
> > done
> us-central1
> us-west1
> ```

@BenTheElder, fascinating.
It appears to have a different result for 
```bash
for REGION in "${REGIONS[*]}"; do 
  echo "$REGION"
done
```

It appears my understanding of array evaluation with loops and expanding of those values is mistaken. I tend to use `${THING[*]}` more often and it seems that `${THING[@]}` is more useful. Will look into why I ended up using asterisks, there was something!

/close

BenTheElder:
Yes -- asterisk and @ are different, @ preserves array behavior and asterisk concatenates with the separator.

https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html

> ${!name[@]}
> ${!name[*]}
>
> If name is an array variable, expands to the list of array indices (keys) assigned in name. If name is not an array, expands to 0 if name is set and null otherwise. When ‘@’ is used and the expansion appears within double quotes, each key expands to a separate word.

BenTheElder:
https://www.gnu.org/software/bash/manual/bash.html#Arrays is another good page

BobyMCbobs:
Thank you, @BenTheElder !

PR #60, opened 2022-04-21
improve logging

### LINKED ISSUE #58: increase log level in sandbox deployment
we should log lots of stuff in the sandbox so we can verify what is happening, but not in prod with the scale of requests we expect there.

we already have klog with verbosity levels, let's increase it a bit in the sandbox deployment, and then add some more verbose-only logging.

BenTheElder:
fixes #58

### DIFF
--- a/cmd/archeio/app/handlers.go
+++ b/cmd/archeio/app/handlers.go
@@ -61,12 +61,12 @@ func makeV2Handler(upstreamRegistry string) func(w http.ResponseWriter, r *http.
 	// capture these in a http handler lambda
 	return func(w http.ResponseWriter, r *http.Request) {
 		path := r.URL.Path
-		klog.V(2).InfoS("v2 request", "path", path)
 
 		// check if blob request
 		matches := reBlob.FindStringSubmatch(path)
 		if len(matches) != 2 {
 			// doesn't match so just forward it to the main upstream registry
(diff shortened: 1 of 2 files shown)


BenTheElder:
I can see these logs in the us-west1 sandbox instance after running `crane pull registry-sandbox.k8s.io/pause:3.1 /dev/null`

Started another run of https://prow.k8s.io/?job=e2e-kops-grid-gcr-mirror-canary so we can see the AWS behavior, those show up in the us-central1

BenTheElder:
Looks like we are *not* redirect to S3.

 > Info
2022-04-21 14:16:19.091 PDTGET308918 B1 mscontainerd/v1.6.2 https://registry-sandbox.k8s.io/v2/pause/manifests/sha256:c2280d2f5f56cf9c9a01bb64b2db4651e35efd6d62a54dcfc12049fe6449c5e4
> Default
2022-04-21 14:16:21.069 PDTI0421 21:16:21.069620 1 handlers.go:85] "redirecting blob request to upstream registry" path="/v2/pause/blobs/sha256:6270bb605e12e581514ada5fd5b3216f727db55dc87d5889c790e4c760683fee"

Though also a lot of requests are coming from cloud CDN which I think we need to disable. I found out from @ameukam today we have this on the LB.

BenTheElder:
https://github.com/kubernetes-sigs/oci-proxy/pull/61 for debugging

PR #61, opened 2022-04-21
REVERT ME: debug why we're not redirecting kops to S3

BenTheElder:
I suspect we should not stay in the business of logging client IPs even with few people having access.
But this should *only* be deployed to the sandbox instance, which is not advertised to end users, only CI, and something is up there. Need to get some real data to debug. https://github.com/kubernetes-sigs/oci-proxy/pull/60#issuecomment-1105771135

### DIFF
--- a/cmd/archeio/app/handlers.go
+++ b/cmd/archeio/app/handlers.go
@@ -79,6 +79,10 @@ func makeV2Handler(upstreamRegistry string) func(w http.ResponseWriter, r *http.
 			return
 		}
 
+		// TODO: revert this!
+		// we just need some real data to debug
+		klog.InfoS("blob request client info", "clientIP", clientIP, "headers", r.Header, "path", path)
+
 		region, matched := regionMapper.GetIP(clientIP)
 		if !matched {
 			// no region match, redirect to main upstream registry


BenTheElder:
waiting for deploy https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-oci-proxy-push-images/1517259944359890944

BenTheElder:
I made a request myself and what I'm seeing is that `X-Forwarded-For` is like:

> X-Forwarded-For:[$my_ip, 2600:1901:0:a7aa::,$my_ip]


`2600:1901:0:a7aa::` [is the registry-sandbox.k8s.io public IP](https://github.com/kubernetes/k8s.io/blob/af102646a1678172e2a5c837bc5c6745d055abe0/dns/zone-configs/k8s.io._0_base.yaml#L256).

but the GCP docs https://cloud.google.com/load-balancing/docs/https#x-forwarded-for_header

have the following variants documented:
> `X-Forwarded-For: <client-ip>,<load-balancer-ip>`
> `X-Forwarded-For: <supplied-value>,<client-ip>,<load-balancer-ip>`

??? cloud run magic inverting these?

BenTheElder:
https://www.reddit.com/r/googlecloud/comments/rbsodd/xforwardedfor_from_external_load_balancer_shows_a/?utm_source=share&utm_medium=ios_app&utm_name=iossmf

BenTheElder:
Commented on b/209919936 internally about this.

In the meantime one option is the standardized [RFC 7239](https://datatracker.ietf.org/doc/html/rfc7239) `Forwarded` header which GCLB does not mention in the docs but does appear to set.

I think this header also has a minor RFC violation in that IPv6 values are not enclosed in brackets, but that seems more reasonable to handle parsing than values in entirely the wrong order. We also do not need to worry about order as the header is key=value based, it will just be more complex to parse ...

BenTheElder:
https://github.com/knative/serving/blob/ebdd09e022eba9697e174065474fc8c486df3f69/test/conformance/runtime/header_test.go#L38

`Forwarded` header is part of knative conformance and Knative shims it when only `X-Forwarded-*` headers are set.
So we can reasonably expect `Forwarded` to remain available in cloud run.

BenTheElder:
You can inject into both of them from the client though (with the trusted values appended), so either way you need to walk backwards and know how many levels to skip (for us currently there's just one forwarder, the LB).

BenTheElder:
What's extra odd is while it says:

> `X-Forwarded-For: <supplied-value>,<client-ip>,<load-balancer-ip>`

When I send a value, what I actually see is:

> `X-Forwarded-For: <supplied-value>, <client-ip>, <load-balancer-ip>,<client-ip>`

BenTheElder:
revert: https://github.com/kubernetes-sigs/oci-proxy/pull/65

PR #66, opened 2022-04-26
HEAD check and fallback if S3 is missing layers

BenTheElder:
The test bucket we have is a one-off sync and cannot be guaranteed to have all layers present.

This caused issues with our test job right after implementing redirect-to-s3 when kops started using an image we didn't have synced into the test s3 bucket. 
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/e2e-kops-grid-gcr-mirror-canary/1517823558204526592

... so for the sake of development, I implemented checking if the layer actually exists.

Each AWS bucket can handle 5,500s HEAD RPS without sharding the bucket key prefix, if we need more we can implement key prefix, or we can institute caching in the OCI proxy. There appears to be no client rate limit, just the bucket scaling.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html

If for any reason we can't verify that the blob is available, we simply fall back to redirecting GCR.
This logic only impacts the AWS-client serving path, and AWS clients already get the benefit of moving to being served a local copy when we roll this out, which should be a net-win.

### DIFF
--- a/cmd/archeio/app/buckets.go
+++ b/cmd/archeio/app/buckets.go
@@ -16,10 +16,41 @@ limitations under the License.
 
 package app
 
-func regionToBucket(region string) string {
+import "net/http"
+
+// awsRegionToS3URL returns the base S3 bucket URL for an OCI layer blob given the AWS region
+//
+// blobs in the buckets should be stored at /containers/images/sha256:$hash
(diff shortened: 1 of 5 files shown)


dims:
/approve
/lgtm

thanks @BenTheElder !

PR #67, opened 2022-04-26
cache bucket HEAD requests

BenTheElder:
follow up to #66 

NOTE: this is a very naive (but should be reasonably fast) cache with **no eviction**. This is fine because we've selected 10 buckets https://github.com/kubernetes-sigs/oci-proxy/issues/39#issuecomment-1085230292 and there are approximately 97,000 blobs, with hashes being 64 bytes we can cache in approximately 10 * 100,000 * 64 bytes AKA 64 MB *even if* we actually observe _all_ keys in _all_ buckets for the lifecycle of the container, which is not remotely likely (consider e.g. that requests should be routed to the nearest cloud run instance, in addition to number of containers being scaled up/down).

### DIFF
--- a/cmd/archeio/app/buckets.go
+++ b/cmd/archeio/app/buckets.go
@@ -16,7 +16,10 @@ limitations under the License.
 
 package app
 
-import "net/http"
+import (
+	"net/http"
+	"sync"
+)
 
 // awsRegionToS3URL returns the base S3 bucket URL for an OCI layer blob given the AWS region
 //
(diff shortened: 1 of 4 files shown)


BenTheElder:
see also discussion: https://kubernetes.slack.com/archives/CCK68P2Q2/p1651013907645179?thread_ts=1651013625.418819&cid=CCK68P2Q2

- cloud run gets 512MiB by default, we're not using a whole lot else (basic request handler overhead/garbage, a little for the CIDR map)
- we can configure more memory if we need (we really shouldn't)
- we should probably bother to decode the hex hash to bytes and save some extra memory that way at the minor cost of decoding time
- an LRU would be appropriate, but existing libs don't appear well-maintained and are not type-safe, hoping one comes along so we don't have to reinvent all the wheels or lose type safety (and the performance lost of cast to interface{})

PR #68, opened 2022-04-26
log cache hits

BenTheElder:
... so we can more easily determine if this is occurring

I see some very fast (1ms) redirect-to-s3 handling occurring sometimes, but let's make it clear.


hack/make-rules/deploy.sh
```diff
@@ -35,5 +35,5 @@ for REGION in "${REGIONS[@]}"; do
         --image "${IMAGE_REPO}:${TAG}" \
         --region "${REGION}" \
         `# TODO: if we use this to deploy prod, we need to handle this differently` \
-        --args=-v=2
+        --args=-v=3
```
BenTheElder, line 38: v(2) seemed a bit low for this, raising debug deployment to v3.
we probably don't want to totally max this out even in the sandbox, we can do _really_ verbose logs locally if needed

Issue #77, opened 2022-06-22
regionalize GCR in oci-proxy


BenTheElder:
Since k8s.gcr.io regionalizes to {eu,us,asia} GCR registries today, *if* we have the GCR team help us reroute the traffic to this domain to registry.k8s.io / oci-proxy, then we will need to regionalize ourselves.

With that team and then with the community in the SIG k8s infra meeting we've discussed the approach of simply doing:
- GCLB routes to the local cloud run instance
- local cloud run instance introspects region or has per-region config to identify the nearest GCR/AR registry

This should be pretty straightforward, but we should look into this a bit more before moving forward.

BenTheElder:
AIUI cloud-run services are regional, we have configured a loadbalancer in front of N regional services.

So we should be able to do this without any code changes actually, just setting `UPSTREAM_REGISTRY` env per each service / region.

ameukam:
> just setting UPSTREAM_REGISTRY env per each service / region.

The current HCL code handling the cloud-run services don't really allow to do that. There is only one HCL construct to handle all the cloud-run services: 

https://github.com/kubernetes/k8s.io/blob/58af71b38de3493b76b0440dff817e07f68795c9/infra/gcp/terraform/k8s-infra-oci-proxy-prod/oci-proxy-prod.tf#L82-L121

 We need to split between one construct per continent or construct per region.

BenTheElder:
ACK -- that's fine, the point is that it's technically feasible and actually doesn't even require additional complexity in the runtime code, it will definitely add a touch to the deployment config.

BenTheElder:
actually to clarify: it will need code change, because the API v2 endpoint is at us.gcr.io/v2 even though the registry is us.gcr.io/k8s-artifacts-prod

we will have to remap the URLs more intelligently than just substituting the base URL.

/assign

BenTheElder:
https://github.com/kubernetes-sigs/oci-proxy/pull/78 solves the auth problem with remapping registry.k8s.io/foo => us.gcr.io/k8s-artifacts-prod/foo

fairly minor code tweaks can inject the "k8s-artifacts-prod" aspect when needed.
I've hacked this up to prove that this issue is a viable approach, but I'll need to follow up with a clean PR.

so far I broke off the auth fix the other day, I'll follow up with the URL rewriting O(soon).

BenTheElder:
Thanks.

[...]

PR #78, opened 2022-06-25
serve 200 OK for /v2/ API check

BenTheElder:
part of #77 

I think we could instead serve 401 and craft an appropriate `Www-Authenticate: Bearer realm` header that includes the scope (unlike GCR, which lets the client determine it based on the repo it parses in the image name), but that's more complex and unnecessary.

Our target use case involves publicly readable registries, and GCR public read is *really* public read, you don't need to present an auth token for any of the API calls.

### DIFF
--- a/cmd/archeio/app/handlers.go
+++ b/cmd/archeio/app/handlers.go
@@ -45,7 +45,7 @@ func MakeHandler(upstreamRegistry string) http.Handler {
 		// v1 API is super old and not supported by GCR anymore.
 		path := r.URL.Path
 		switch {
-		case strings.HasPrefix(path, "/v2/"):
+		case strings.HasPrefix(path, "/v2"):
 			doV2(w, r)
 		case path == "/":
 			http.Redirect(w, r, infoURL, http.StatusPermanentRedirect)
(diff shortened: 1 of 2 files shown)


cmd/archeio/app/handlers.go
```diff
@@ -41,11 +41,17 @@ func MakeHandler(upstreamRegistry string) http.Handler {
 	blobs := newCachedBlobChecker()
 	doV2 := makeV2Handler(upstreamRegistry, blobs)
 	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// this registry only supports proxy image downloads / listing
+		// we do *not* want to deal with mutations, those are done against
+		// the backing registries by other tools
+		if r.Method != http.MethodGet && r.Method != http.MethodHead {
```
BenTheElder, line 47: unrelated tightening of the handler that I perhaps should have moved to another commit.

all of the endpoints we should support use one of these two methods https://github.com/opencontainers/distribution-spec/blob/main/spec.md#endpoints

POST/PUT/PATCH/DELETE are in the spec as well, but of course used for mutations (pushing images etc.), which we don't need or want to support here

BenTheElder: this needs test coverage, perhaps punt to another PR, it's not important.

ameukam:
/lgtm
/hold

Remove the hold when ready.

PR #82, opened 2022-06-30
Update AWS region to S3 url to use production buckets

BobyMCbobs:
Once the buckets have been synced, use the production S3 bucket URLs

Depends on: https://github.com/cncf-infra/aws-infra/issues/4, https://github.com/kubernetes/k8s.io/issues/3623

### DIFF
--- a/cmd/archeio/app/buckets.go
+++ b/cmd/archeio/app/buckets.go
@@ -17,6 +17,7 @@ limitations under the License.
 package app
 
 import (
+	"fmt"
 	"net/http"
 	"sync"
 
@@ -28,6 +29,8 @@ import (
 // blobs in the buckets should be stored at /containers/images/sha256:$hash
 func awsRegionToS3URL(region string) string {
 	switch region {
+	case "us-west-1", "us-west-2", "us-east-1", "us-east-2", "eu-central-1", "ap-southeast-1", "ap-northeast-1", "ap-south-1":
+		return fmt.Sprintf("https://prod-registry-k8s-io-%v.s3.%v.amazonaws.com", region, region)
 	default:
 		return "https://prod-registry-k8s-io-us-east-2.s3.dualstack.us-east-2.amazonaws.com"
 	}
--- a/cmd/archeio/app/buckets.go
+++ b/cmd/archeio/app/buckets.go
@@ -29,7 +29,7 @@ import (
 // blobs in the buckets should be stored at /containers/images/sha256:$hash
 func awsRegionToS3URL(region string) string {
 	switch region {
-	case "us-west-1", "us-west-2", "us-east-1", "us-east-2", "eu-central-1", "ap-southeast-1", "ap-northeast-1", "ap-south-1":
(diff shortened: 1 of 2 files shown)


cmd/archeio/app/buckets.go
```diff
@@ -29,8 +30,11 @@ import (
 func awsRegionToS3URL(region string) string {
 	// for now always return @ameukam's test bucket
```
sftim, line 31: Maybe remove this comment?

BobyMCbobs: Thanks for noticing it, I've removed it in b78adbe

cmd/archeio/app/buckets.go
```diff
@@ -27,10 +28,12 @@ import (
 //
 // blobs in the buckets should be stored at /containers/images/sha256:$hash
 func awsRegionToS3URL(region string) string {
-	// for now always return @ameukam's test bucket
 	switch region {
+	case "us-west-1", "us-west-2", "us-east-1", "us-east-2", "eu-central-1", "ap-southeast-1", "ap-northeast-1", "ap-south-1":
+		return fmt.Sprintf("https://prod-registry-k8s-io-%v.s3.%v.amazonaws.com", region, region)
```
ameukam, line 33:
```suggestion
		return fmt.Sprintf("https://prod-registry-k8s-io-%v.s3.dualstack.%v.amazonaws.com", region, region)
```

cmd/archeio/app/buckets.go
```diff
@@ -27,10 +28,12 @@ import (
...
 	default:
-		return "https://painfully-really-suddenly-many-raccoon-image-layers.s3.us-west-2.amazonaws.com"
+		// default to us-east-2 for stability purposes
+		return "https://prod-registry-k8s-io-us-east-2.s3.us-east-2.amazonaws.com"
```
ameukam, line 36:
```suggestion
		return "https://prod-registry-k8s-io-us-east-2.s3.dualstack.us-east-2.amazonaws.com"
```

sftim, line 36: This could surprise a caller who expects the function to succeed for any valid AWS region.

I'd be tempted to have a single format string in the function that is always used on the happy parh, or fail the lookup if we don't recognize the region (and then, callers must check for an error).

In case the region isn't recognized:
We can have a different function eg `awsFallbackS3Url` that returns a constant value, or we can serve from elsewhere.

cmd/archeio/app/buckets.go
BobyMCbobs: Updated in 7a74fa6

cmd/archeio/app/buckets.go
BobyMCbobs: Updated in 7a74fa6

BobyMCbobs:
All the buckets appear to be synced manually. This is ready!

/unhold

BobyMCbobs:
Noticed there was a missing bucket in https://github.com/kubernetes-sigs/oci-proxy/pull/82/commits/5d990df1a4b7298943bb39e76ad8148eb821bd23

jaypipes (approved):
All looks sensible to me. :+1:

BobyMCbobs:
Thank you @dims!

I think we're good to go!
/unhold

PR #86, opened 2022-07-02
Add endpoint to serve LF privacy page

### LINKED ISSUE #81: serve privacy policy
We determined in #62 that we should be operating under the linux foundation privacy policy

However, archeio does not itself serve the privacy policy in any way.

We should add an endpoint for humans like `/privacy` (this is fine, registry API calls are all under `/v2/`) that redirects to the LF privacy page to start.

It might be worth considering a landing page with some context and a link to the LF page in the future instead.

jeremyphua:
Fixes #81 

## Proposed changes

- Add `/privacy` endpoint  to `handlers.go` to serve LF privacy policy page.
- Add new endpoint to test case in `handlers_test.go`.
- Update request-handling.md text and chart flow.

### DIFF
--- a/cmd/archeio/app/handlers.go
+++ b/cmd/archeio/app/handlers.go
@@ -28,7 +28,8 @@ import (
 
 const (
 	// hard coded for now
-	infoURL = "https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io"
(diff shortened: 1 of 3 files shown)


ameukam:
/lgtm
/approve

Thanks for doing this!

PR #87, opened 2022-07-04
Enable serving Arbitary Registries

upodroid:
Fixes: #77

This PR enables oci-proxy to serve images from GCR and AR repositories.

Simple demo of bug:
```
# oci-proxy is running in my project with UPSTREAM_REGISTRY set to https://eu.gcr.io/k8s-artifacts-prod.
 REDACTED  MCW0CDP3YY  ~  Desktop  Git  oci-proxy   regionalise-gcr  1+  1⚑  SIGINT  $   curl https://oci-proxy-4txm7cz3ca-ew.a.run.app/v2/pause/manifests/latest -IL -XGET
HTTP/2 308
content-type: text/html; charset=utf-8
location: https://eu.gcr.io/k8s-artifacts-prod/v2/pause/manifests/latest
x-cloud-trace-context: 34a1d87b5ef3817f23c6e4b797cdaaa3;o=1
date: Mon, 04 Jul 2022 11:56:32 GMT
server: Google Frontend
content-length: 98
alt-svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"

HTTP/2 302
location: https://console.cloud.google.com/m/gcr/images/redirect?project=k8s-artifacts-prod&location=EU&repo_name=v2/pause/manifests/latest
content-length: 334
date: Mon, 04 Jul 2022 11:56:32 GMT
content-type: text/html; charset=UTF-8
server: Docker Registry
x-xss-protection: 0
x-frame-options: SAMEORIGIN
alt-svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000,h3-Q050=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000,quic=":443"; ma=2592000; v="46,43"

```

After this PR:

```
# UPSTREAM_REGISTRY=https://europe-docker.pkg.dev/v2/coen-mahamed-ali/images ./bin/archeio -v=9
# This is an AR backing registry with a complete copy of gcr.io/k8s-artifacts-prod

 REDACTED  MCW0CDP3YY  ~  Desktop  Git  oci-proxy   regionalise-gcr  3✎  1+  1⚑  $   curl http://localhost:8080/v2/pause/manifests/latest -iL -XGET
HTTP/1.1 308 Permanent Redirect
[...]

PR #88, opened 2022-07-04
Feature gate serving AWS S3 blobs from oci-proxy

### LINKED ISSUE #80: gate or temporarily compile out s3 redirect
since the s3 buckets are still not ready, we should not block #77 rollout to registry.k8s.io on production s3 bucket availability.

I didn't think it would take this long, so I went with the simpler option of not gating s3 redirection / maintaining two modes, figuring we'd promote to production soon enough, and the initial production app is *literally* just bog simple 3XX to k8s.gcr.io so we can get users on registry.k8s.io and shouldn't need patches.

since it is taking a while, we should make it possible to continue iterating on other aspects, like redirecting to the individual GCR regions backing k8s.gcr.io ourselves, and later artifact registry instances instaed.

upodroid:
Fixes #80

/cc @dims @ameukam 

~~Arnaud, we need to modify the sandbox instance of oci-proxy and set `SERVE_IMAGES_FROM_AWS` to `true`~~

### DIFF
--- a/cmd/archeio/app/handlers.go
+++ b/cmd/archeio/app/handlers.go
@@ -31,6 +31,7 @@ type RegistryConfig struct {
 	UpstreamRegistryPath     string
 	InfoURL                  string
 	PrivacyURL               string
+	ServeImagesfromAWS       bool
 }
 
 // MakeHandler returns the root archeio HTTP handler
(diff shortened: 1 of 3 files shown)


ameukam:
This may conflict with the purpose of https://github.com/kubernetes-sigs/oci-proxy/pull/82. 

Let's get feedback from folx working the AWS part before we move forward with this.

upodroid:
I can switch the default to true instead of false and amend the environment variables for the prod deployment if the buckets aren't ready by then.

ameukam:
> I can switch the default to true instead of false and amend the environment variables for the prod deployment if the buckets aren't ready by then.

In that case we should introduce this after we start serve from the S3 buckets. There is not specific value to introduce logic that already default the existing behavior.

/hold

dims:
@upodroid comments for 2 minor things left inline, PTAL

upodroid:
This can be merged now. Rebased with the changes from #87 and serving from AWS is automatically turned on.

BobyMCbobs:
@upodroid, thanks for your changes! It looks fine.

My only question is: since the blobs are now synced, are we still needing to feature gate the S3 buckets?

Ref: #82

BobyMCbobs:
> Yes, it is useful for anyone looking to use archeio as an image proxy.
> Also, if we ever have S3 issues, we have an easy way of turning it off in production.

@upodroid, sounds good to me!

/lgtm

jaypipes (approved):
Yeah, seems reasonable to me. ++

sftim (commented):
Informally: LGTM

cmd/archeio/app/handlers.go
```diff
@@ -97,47 +98,49 @@ func makeV2Handler(rc RegistryConfig, blobs blobChecker) func(w http.ResponseWri
...
-
-		// for blob requests, check the client IP and determine the best backend
-		clientIP, err := getClientIP(r)
-		if err != nil {
-			// this should not happen
-			klog.ErrorS(err, "failed to get client IP")
-			http.Error(w, err.Error(), http.StatusBadRequest)
-			return
-		}
+		if rc.ServeImagesfromAWS {
```
BenTheElder, line 101: TODO(minor): this is only set at startup and we're returning a request handler function, we can avoid the runtime check on each request.

Issue #83, opened 2022-07-01
consider a more useful `/` redirect


BenTheElder:
http://registry-sandbox.k8s.io/ redirects to https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io which doesn't actually contain anything we're using or any docs.

https://github.com/kubernetes/k8s.io/wiki/New-Registry-url-for-Kubernetes-(registry.k8s.io) is probably a better URL for now?

cc @dims @ameukam 

We should also probably apply #79 to paths like this, since we'll update it in the future.

BenTheElder:
Ordinarily for `help wanted` issues the author would probably help guide contributors, but I am going on vacation for a few weeks and will be relatively unavailable until sometime after July 25th.

`@ameukam` and `@dims` are repo approvers / admins and should be able to help in the meantime.

See also https://kubernetes.slack.com/archives/CCK68P2Q2/p1656629484801139?thread_ts=1656626472.238269&cid=CCK68P2Q2 in Kubernetes slack.

---

For this issue, you can find the http handler in cmd/archeio/app/handlers.go

You'll also have to update the test cases.

BenTheElder:
Accidentally reverted in https://github.com/kubernetes-sigs/oci-proxy/pull/87#discussion_r929470941, we should switch this again

Issue #72, opened 2022-05-04
Mapping AWS regions and S3 bucket region


ameukam:
Part of:
  - https://github.com/kubernetes/k8s.io/issues/3411
  
From https://github.com/kubernetes-sigs/oci-proxy/issues/39, we identified which region the traffic come the most but we need to ensure redirection for traffic coming from regions not listed. 
We could also assume the rest of regions are served from only 1 bucket but they is a risk of high latency during image pulling.

Draft:

| regions | bucket name | bucket location |
| -------- | -------- | -------- |
| us-east-1, sa-east-1     |      | us-east-1     |
| ca-central-1, us-east-2     |      | us-east-2     |
| us-west-1     |      | us-west-1     |
| us-west-2     |      | us-west-2     |
| ap-south-1, me-south-1     |      | ap-south-1    |
| ap-northeast-1, ap-northeast-2, ap-northeast-3     |      | ap-northeast-1 |
| ap-east-1, ap-southeast-1, ap-southeast-2. ap-southeast-3     |      | ap-southeast-1|
| eu-central-1, eu-south-1     |      | eu-central-1     |
| eu-west-1, af-south-1    |      | eu-west-1     |
| eu-north-1, eu-west-2, eu-west-3     |      | eu-west-2     |

/sig k8s-infra
/area artifacts
/priority important-soon
/milestone v1.25
/assign @BenTheElder

BenTheElder:
cc @jaypipes 

> Not listed:
> ap-southeast-3
> ap-northeast-3
> ap-southeast-2
> sa-east-1

I would guess:
- ap-southeast-3 => ap-southeast-1
- ap-northeast-3 => ap-northeast-1
- ap-southeast-2 => ap-southeast-1
- sa-east-1 => us-east-1

Riaankl:
> > cc @jaypipes
> > > Not listed:
> > > ap-southeast-3
> > > ap-northeast-3
> > > ap-southeast-2
> > > sa-east-1
> > 
> > 
> > I would guess:
> > ```
> > * ap-southeast-3 => ap-southeast-1
> > 
> > * ap-northeast-3 => ap-northeast-1
> > 
> > * ap-southeast-2 => ap-southeast-1
> > 
> > * sa-east-1 => us-east-1
> > ```
> 
> @BenTheElder yep, that works. +1
> 
> Keep in mind the amount of traffic coming from these regions is very low.

Looking at the Data Studio report all four regions <2% total

[...]

PR #92, opened 2022-07-26
Fix infoURL

upodroid:
Fixes: #83 

/cc @BenTheElder 

How would I go about optimising https://github.com/kubernetes-sigs/oci-proxy/pull/88/files#r929468837 ?


BenTheElder (approved):
/lgtm
/approve

thanks!

BenTheElder:
> How would I go about optimising https://github.com/kubernetes-sigs/oci-proxy/pull/88/files#r929468837 ?

We can just return a different version of the function depending on if we're serving from AWS.

Given that the buckets appear to be up now though, to be honest ... I have half a mind to suggest we keep the rest of the refactoring from that PR but drop gating AWS soon and just focus on rolling out AWS, promptly. For that I think we're just missing implementing the region to bucket mapping #72, we have the code in place, we just need to identify the bucket names and put them into the mapping.

BenTheElder:
Long term less code is better here, we'll want to be sure it is all carefully tested and reviewed and highly trusted given that we'll sit in front of image pulls (!)

PR #93, opened 2022-08-06
Ensure deployment to Tier1 regions

ameukam:
Ensure oci-proxy will be deployed to the Tiers 1 regions of Cloud Run


BobyMCbobs:
The given regions appear to also exist in Cloud Run
image

which means this should run just fine! 

/lgtm

PR #95, opened 2022-08-08
Refactor Terraform Code

upodroid:
/cc @ameukam @dims @BenTheElder 

1. This PR replaces https://github.com/kubernetes/k8s.io/pull/3904
2. We need to stop deploying the sandbox instance as part of the post-oci-proxy-push-images postsubmit job.
3. A new presubmit terraform job that runs plan. To block running untrusted code, the terraform script will be embedded in the prow job 
4. A new postsubmit terraform job that runs plan for dev and prod. To block running untrusted code, the terraform script will be embedded in the prow job.
5. A generic image bumper job to bump the images in images.yaml file.

From Original PR:

_I rewrote the terraform code to allow per region modification of each cloud run service now. Right now, you can specify env variables for each specific region but other values can be changed later too._

_It still needs more improvements but i'll need privileges to execute those changes. For example, you should be using project-factory module._


.terraform-version
```diff
@@ -0,0 +1 @@
+1.1.0
```
ameukam, line 1: why do we need this ?

.gitignore
```diff
@@ -11,3 +11,7 @@
 *.sublime-project
 *.sublime-workspace
 *~
+
+# terraform
+.terraform
+terraform.tfstate
```
ameukam, line 17:
```suggestion
# terraform
.terraform
terraform.tfstate
.terraform.lock.hcl
```

terraform/sandbox/.terraform.lock.hcl
```diff
@@ -0,0 +1,59 @@
+# This file is maintained automatically by "terraform init".
```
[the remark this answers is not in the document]
upodroid: It is recommended to commit that. Like go.mod.

https://www.terraform.io/language/files/dependency-lock#lock-file-location

.terraform-version
upodroid: I'll delete this.

BenTheElder:
> We need to stop deploying the sandbox instance as part of the post-oci-proxy-push-images postsubmit job.
>
> [...]
>
> A new postsubmit terraform job that runs plan for dev and prod. To block running untrusted code, the terraform script will be embedded in the prow job.


I think we should handle this part a little differently ...

We can run whatever we want in that job because it's a cloudbuild.yaml in this repo, without removing the job.

We need to do staging rollouts *after* we push the image or else give up autodeploying latest to staging.

We can instead update the cloudbuild.yaml to run the terraform for the deployment step.

We probably should not deploy prod and staging from the same job, so prod can rollout faster without blocking on the image push (prod shouldn't immediately adopt new images anyhow) and so we can further isolate the credentials.

We can add new cloudbuild postsubmit for prod only.

We also should not rely on embedding the script in an image to run in prow, the "trusted" prow cluster is the only secure cluster and we very very rarely approve additional new code sources running directly in that cluster, because it's a security nightmare.

Instead, we permit projects to submit a prow job that launches a cloudbuild in their own GCP project, using a small fixed tool owned by the same reviewers. This is much easier to review safely, and ensures that each project's trusted workloads are in an isolated GCP project with their own credential management.

terraform/sandbox/.terraform.lock.hcl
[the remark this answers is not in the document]
ameukam: This is an ongoing issue about lock files on different platforms: https://github.com/hashicorp/terraform/issues/28041.

We should apply the workaround suggested  (generated the lock file for all platforms) or remove entirely the file. Also the lock file is not really a requirement to deploy changes like the `go.sum` file; I could just remove the file and still be capable to run `terraform apply`.

ameukam:
> We probably should not deploy prod and staging from the same job, so prod can rollout faster without blocking on the image push (prod shouldn't immediately adopt new images anyhow) and so we can further isolate the credentials. 

There is an extra step of image promotion for `archeo` before the rollout to prod.
@upodroid I think for the moment, we don't want the production HCL code in this repository. we should look on how to trigger production deployment after image promotion.

BenTheElder:
> There is an extra step of image promotion for archeo before the rollout to prod.

That's not related to the production config rollout. Deciding to roll out a new image is not necessarily happening in all config changes.

But in all commits to this repo, we should be rolling out a new image to staging, so *that* config change should probably happen after the image is pushed. Whereas production config patches should not be waiting on the staging image build ...

dims:
@upodroid looks like there's a bunch of feedback in here?

PR #96, opened 2022-08-11
Bump go version

ameukam:
Bump go version and components used for tests


ameukam:
cc @BenTheElder

BenTheElder:
there's a new lint failure. https://github.com/kubernetes-sigs/oci-proxy/pull/21/files needs to be extended to ReadHeaderTimeout

.go-version
```diff
@@ -1 +1 @@
-1.18.1
+1.18.5
```
BenTheElder, line 1: why not 1.19?

ameukam: wait for 1.19.1 ? I don't have a strong opinion but I'll prefer we pick .1 for every major version.

BenTheElder: Kubernetes doesn't wait 🤷‍♂️ 

All minor go releases are actually soaked in google prod before release, and we should be much less sensitive to go changes than Kubernetes ... we don't have 5k node scale issues or high performance requirements.

ameukam: Bumped.

BenTheElder (approved):
/lgtm
/approve

ameukam:
`gimme` is currently not supporting 1.19 (not working with `1.19`). I'll switch to `1.19` when https://github.com/travis-ci/gimme/pull/203 is merged. 

cc @dims

BenTheElder (commented):
/shrug
/lgtm
/approve

BenTheElder:
gimme does not need to know that a go version exists to download it, you just need to make sure to use 1.19 not 1.19.0 (1.19.0 is not a go version) https://github.com/kubernetes-sigs/kind/blob/main/hack/third_party/gimme/gimme

PR #99, opened 2022-09-01
update ip addresses

ameukam:
Update ip addresses from AWS IP ranges.
Also shebang to better support bash installation on other platforms.


BenTheElder:
In the future I recommend generated changes in a separate commit, but the non-generated changes are small and the total number of files is smile so it's fine 😅

BenTheElder (approved):
/lgtm
/approve

ameukam:
> In the future I recommend generated changes in a separate commit, but the non-generated changes are small and the total number of files is smile so it's fine 😅 

Possibly, a periodic generating the changes for us?

PR #103, opened 2022-09-15
Copy Prod TF Code

upodroid:
This PR **just copies** the code from k/k8s.io. I will make the changes in a followup PR.

/cc @BenTheElder @ameukam


ameukam:
The HCL code for production workloads should only be in https://git.k8s.io/k8s.io. we should only HCL code for sandbox here.

PR #108, opened 2022-09-28
update tools dependencies

BenTheElder:


go.mod
```diff
@@ -1,6 +1,6 @@
 module k8s.io/registry.k8s.io
 
-go 1.18
+go 1.19
```
BenTheElder, line 3: we don't care about supporting older go, this repo is only used by us and built with the latest go.

hack/tools/go.mod
```diff
@@ -1,16 +1,240 @@
...
 
-go 1.13
+go 1.19
 
 require (
-	github.com/go-xmlfmt/xmlfmt v0.0.0-20211206191508-7fd73a941850 // indirect
-	// temporary until https://github.com/golangci/golangci-lint/issues/2649
-	// is resolved, then we will go back to released versions
-	github.com/golangci/golangci-lint v1.48.0
+	github.com/golangci/golangci-lint v1.43.0
```
BenTheElder, line 6: eep this is older

PR #114, opened 2022-10-11
add some background details, and a note about privacy.

BenTheElder:
this is essentially a TLDR of the motivation from the existing doc, and a mention of the privacy policy.

this should let us start to redirect registry.k8s.io/ to the current repo docs instead of the "new url" doc (which captures the registry move as planned in a point in time, but may fall out of date more easily)

next up: change the `/` redirect in cmd/archeio

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,34 @@
 This project implements the backend for registry.k8s.io, Kubernetes's container
 image registry.
 
-For more see [cmd/archeio](./cmd/archeio/README.md)
+For more details on the implementation see [cmd/archeio](./cmd/archeio/README.md)
+
+The community deployment configs are documented at in the k8s.io repo with
+the rest of the community infra deployments:
+https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io
+
+## Privacy
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -3,7 +3,34 @@
...
+The community deployment configs are documented at in the k8s.io repo with
+the rest of the community infra deployments:
+https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io
+
+## Privacy
+
+This project abides by the Linux Foundation privacy policy, as documented at
+https://registry.k8s.io/privacy
+
+## Background
```
BenTheElder, line 17: this is a much shorter version, to capture some context knowing users have link fatigue, it is not intended to replace the communication about why we made the move wholesale, but to be sufficient that we can point users to this repo with other current details at the `/` redirect

BenTheElder: (and to consolidate current documentation of what exactly registry.k8s.io _is_, we also now have inbound links from https://github.com/kubernetes/k8s.io/tree/main/registry.k8s.io and hopefully soon https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io)

PR #118, opened 2022-10-13
map to nearest AWS regions as previously proposed

BenTheElder:
This ensure we're not routing traffic from Asia to the US ...

The mapping is from https://github.com/kubernetes/registry.k8s.io/issues/72#issue-1225709110

Initial commit _should fail_ CI, because that mapping is actually missing regions. Will follow up with adding those regions.

### DIFF
--- a/cmd/archeio/app/handlers_test.go
+++ b/cmd/archeio/app/handlers_test.go
@@ -160,6 +160,7 @@ func TestMakeV2Handler(t *testing.T) {
 			"https://prod-registry-k8s-io-us-east-2.s3.dualstack.us-east-2.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
 			"https://prod-registry-k8s-io-us-west-1.s3.dualstack.us-west-1.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
 			"https://prod-registry-k8s-io-us-west-2.s3.dualstack.us-west-2.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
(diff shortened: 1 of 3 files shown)


dims:
looks like we have some unit test failures

BenTheElder:
> looks like we have some unit test failures

Yeah .. there are a LOT of AWS regions, the proposed table was not comprehensive. Second commit adds remaining regions but I think there's one more issue

cmd/archeio/app/handlers_test.go
```diff
@@ -160,6 +160,7 @@ func TestMakeV2Handler(t *testing.T) {
 			"https://prod-registry-k8s-io-us-east-2.s3.dualstack.us-east-2.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
 			"https://prod-registry-k8s-io-us-west-1.s3.dualstack.us-west-1.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
 			"https://prod-registry-k8s-io-us-west-2.s3.dualstack.us-west-2.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
+			"https://prod-registry-k8s-io-eu-west-2.s3.dualstack.eu-west-2.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
```
BenTheElder, line 163: we have a test-case with known IP (IP => region tested elsewhere) in eu-west-3, previously that would have been in us-west-1 along with all other unknown regions

so we have to add eu-west-2 test data because that's where we serve eu-west-3

BenTheElder:
Second commit contains all additional region mappings

... I took a best stab at this. I'm not an AWS expert but I can go back and forth between searching regions and comparing the bucket regions on a map. I'm sure I got a few wrong, the local regions were not conveniently listed and it was a bit painful.

Some of these regions we're not going to be serving realistically, like gov-cloud, but it's just easier to not special case them.

The alternate approach is to perhaps just assume a mapping from cloud run region to S3 region and push it out into the deployment config, but this is probably a net improvement that should be good to go for now.

BenTheElder:
tagged v0.0.4 with this + #117

BenTheElder:
https://github.com/kubernetes/k8s.io/pull/4358

PR #119, opened 2022-10-14
drop cloud run bug handling, as the bug has been fixed since july

BenTheElder:
... checked back up on this bug, it's fixed. 

per the issue tracker the fix was done rolling out in july, so we don't need the quirks-mode codepath anymore.


BenTheElder:
will have to keep an eye on staging, we can double check just by seeing if we are still redirecting some traffic to AWS

before we added the quirks mode parsing, all traffic went to GCP and never AWS because the clientIP we parsed was always the loadbalancer due to the bug. should be pretty easy to check if that's happening again (and ... it shouldn't be).

PR #121, opened 2022-10-19
simplify redirect path calculation

BenTheElder:


cmd/archeio/app/handlers.go
```diff
@@ -120,38 +123,30 @@ func makeV2Handler(rc RegistryConfig, blobs blobChecker) func(w http.ResponseWri
...
-		redirectPath := calculateRedirectPath(rc, path)
-		klog.V(2).InfoS("redirecting blob request to upstream registry", "path", path, "redirect", rc.UpstreamRegistryEndpoint+redirectPath)
-		http.Redirect(w, r, rc.UpstreamRegistryEndpoint+redirectPath, http.StatusTemporaryRedirect)
+		redirectURL := upstreamRedirectURL(rc, rPath)
+		klog.V(2).InfoS("redirecting blob request to upstream registry", "path", rPath, "redirect", redirectURL)
+		http.Redirect(w, r, redirectURL, http.StatusTemporaryRedirect)
 	}
 }
 
-func calculateRedirectPath(rc RegistryConfig, path string) string {
```
BenTheElder, line 148: we never call this unless we're redirecting /v2/something to upstream. so we can simplify it a lot.

BenTheElder: we also always wind up adding the upstream endpoint in front of it, so we can just dedupe that.

PR #123, opened 2022-10-28
add a note re: intention for reusability

BenTheElder:


cmd/archeio/README.md
```diff
@@ -23,6 +23,9 @@ Kubernetes SIG K8s-Infra's needs. However, some of the packages under
 [`pkg/`](./../../pkg/) may be useful if you have a similar problem,
 and they should be pretty generalized and re-usable.
 
+We are looking towards making this application easily re-usable by other projects
+in the future, if you're interested please reach out to SIG K8s Infra to discuss.
```
BenTheElder, line 27: We're not there yet. It's not our top priority. But it makes sense in the long run.

I had some good discussions this week about what that direction might look like balancing maintainability, solving our problems effectively, and being easily reused. I intend to write something up when I get a breather.

PR #126, opened 2022-11-02
Simple non-image redirect server

justinsb:
Copying the archeio design for non-image artifacts.  We serve hash
files (and gpg files) directly, everything else can be redirected to a
mirror.


BenTheElder:
still mulling if it's better to have that binary here or create artifacts.k8s.io repo and import pkg/ from here.

for discoverability reasons, the latter is better. (e.g. we redirect https://registry.k8s.io to https://github.com/kubernetes/registry.k8s.io, where the readme explains what this is)

for development reasons, it may be mildly easier to keep them in a single repo.

cmd/archeio/app/buckets.go
```diff
@@ -94,7 +94,7 @@ type cachedBlobChecker struct {
 	blobCache
 }
 
-func newCachedBlobChecker() *cachedBlobChecker {
+func NewCachedBlobChecker() BlobChecker {
```
BenTheElder, line 97: either way, let's keep exported APIs under pkg/

justinsb: Agreed; the changes here are pretty small (just making some things public, and no actual code changes), and I wanted to get agreement before starting to move code around!

justinsb:
> still mulling if it's better to have that binary here or create artifacts.k8s.io repo and import pkg/ from here.

Either is fine with me.  We can always split out to a separate repo later, but at least at the moment it looks like the code is going to stay pretty similar.  The biggest difference is that artifacts chooses serve some files directly (the hashfiles), and _maybe_ registry should be doing that as well for tags.

BenTheElder:
> Either is fine with me. We can always split out to a separate repo later, but at least at the moment it looks like the code is going to stay pretty similar. 

The code will be pretty similar, though the directly reusable portion is probably scoped to the bucket cache and the IP detection utils which we can export in `pkg/` and reuse as a library.

The docs and discoverability are harder though, and the project already renamed this repo once. I think it would be easier to figure that out up front (also if another repo is fine, it won't take long to get one, we just ask k8s infra leads and file a kubernetes/org request).

I lean towards github.com/kubernetes/artifacts.k8s.io, so https://artifacts.k8s.io/ can redirect there and we can also more easily tag releases etc independently.

> The biggest difference is that artifacts chooses serve some files directly (the hashfiles), and maybe registry should be doing that as well for tags.

Archeio only forwards layer downloads to storage other than the upstream backing registry. Tags are a different API call and go to the canonical backing registry, not to s3 etc.

BenTheElder (commented):
given the funding situation, I also imagined wholesale serving all requests from AWS, at least for now, so the initial application for artifacts.k8s.io can be a tiny redirect 1:1 from cloud run region (so the running process / deployment config) to a backing bucket (or a lambda or whatever is appropriate on the AWS side if someone wants to explore that, we can point DNS wherever)

ameukam:
> Copying the archeio design for non-image artifacts. We serve hash files (and gpg files) directly, everything else can be redirected to a mirror.

@justinsb why not use a CDN provider shielding the buckets for any Object Storage service ? Fastly for example allow you to define routing logic based on the ASN number or Cloudfront give the possibility to have multiple S3 buckets as origins.

BenTheElder:
> @justinsb why not use a CDN provider shielding the buckets for any Object Storage service ? Fastly for example allow you to define routing logic based on the ASN number or Cloudfront give the possibility to have multiple S3 buckets as origins.

This specific PR aside, I think it's worth considering a tiny 3xx redirector app similar to, but much less complex than, the registry app. This allows us to flexibly do things in the future like:

- having a single location where we trust sha256 files vs the larger binaries (part of what this PR tries to do)
- swapping out cloud storage for github release page binaries
- serving from copies in different clouds

As for fastly, we ... don't have fastly today, but we do need to cut spend off of kubernetes.io sooner than later.

justinsb:
Handing this over to a CDN changes the security posture also.  Which reminds me... next step here is probably to update the design docs, then the tests :-)

cmd/redirectserver/app/buckets.go
```diff
@@ -0,0 +1,72 @@
...
+	// each of these has the region in which we have a bucket listed first
+	// and then additional regions we're mapping to that bucket
+	// based roughly on physical adjacency (and therefore _presumed_ latency)
+	//
+	// if you add a bucket, add a case for the region it is in, and consider
+	// shifting other regions that do not have their own bucket
+
+	// US East (N. Virginia)
+	case "us-east-1", "sa-east-1", "us-gov-east-1", "GLOBAL":
+		return "https://prod-registry-k8s-io-us-east-1.s3.dualstack.us-east-1.amazonaws.com"
```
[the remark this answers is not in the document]
justinsb: agreed - get me a bucket to point to, I'll update the URLs!

I'll probably just update them all to artifacts.k8s.io for now anyway, just so we _could_ stand this up somewhere if we wanted to.  (Sadly the current bucket is locked down, and signed urls are a bit more painful than I would have hoped)

ameukam:
we should probably move this to https://github.com/kubernetes-sigs/porche

PR #130, opened 2022-11-08
export getClientip as pkg/clientip.Get

BenTheElder:
for reuse by porche.

I almost put this in a package called `cloudrun`, because it's hard to fully generalize this method without making it much more complex and inefficient and it's unclear that we should be prioritizing arbitrary network topologies, versus just implementing specific ones as needed.

currently we assume this is either local HTTP testing in-memory, or else serving in cloud-run, where we can make assumptions about header availability and content.

originally I just didn't export this, to revisit later, but we'll have the same issue in porche

### DIFF
--- a/cmd/archeio/app/handlers.go
+++ b/cmd/archeio/app/handlers.go
@@ -24,6 +24,7 @@ import (
 
 	"k8s.io/klog/v2"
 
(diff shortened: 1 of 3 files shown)


justinsb:
Thanks @BenTheElder, this will be helpful for porche!

/lgtm

PR #132, opened 2022-11-11
clientip: Add tests for IPv6

justinsb:
I thought this was not working correctly because of IPv6, it turned
out it was not working because I was trying to hit the CloudRun
endpoint directly.  Added some test coverage to prove that it works
and document this.

### DIFF
--- a/pkg/clientip/clientip.go
+++ b/pkg/clientip/clientip.go
@@ -30,7 +30,11 @@ import (
 // 1. no loadbalancer, local testing
 // 2. behind Google Cloud LoadBalancer (as in cloudrun)
 //
-// At this time we have no need to complicate it further
+// Note that in particular we do not support hitting the CloudRun endpoint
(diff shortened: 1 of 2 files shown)


pkg/clientip/clientip.go
```diff
@@ -30,7 +30,10 @@ import (
 // 1. no loadbalancer, local testing
 // 2. behind Google Cloud LoadBalancer (as in cloudrun)
 //
-// At this time we have no need to complicate it further
+// Note that in particular we do not support hitting the CloudRun endpoint
+// directly (though we could easily do so).
```
BenTheElder, line 34: Yeah, we should not do support that, no one should know the cloud run endpoint, that would be a security issue. Cloud Armor is on the GCLB.

justinsb: Thanks - updated the note to incorporate the reasoning.

PR #136, opened 2022-12-15
add makefile command to generate the ip ranges & latest updated ranges

ajayk:
- generate-ipranges:  makes it single for both pulling the latest ip ranges and updating the generated files 
- updates to the latest ip range updates from AWS


BenTheElder:
Hi, thanks.
1) make rules should be implemented by a matching script
2) the script should ensure the correct go version
3) we should split this into two rules, one to pull the static data, one to generate generated code (which update should also cover, following the patterns from Kubernetes). I'm also not sure the former should be a top level rule as it's bit of an internal implementation detail subject to change and there's already an easily invoked script specific to this.

Additionally, we do not need to regenerate these very frequently.

It is expected that we do not capture 100% of IPs and we're not deploying new versions to production right now (no other features ongoing and wish to avoid disruptions through the holidays), we can do this infrequently closer to tagging a release.

ameukam:
@ajayk can you update your PR to include only the Makefile rule ? Thanks! 🙏🏾

Makefile
```diff
@@ -94,6 +94,10 @@ update:
 	hack/make-rules/update.sh
 tidy:
 	hack/make-rules/tidy.sh
+# generate ip ranges
+generate-ipranges:
+	pkg/net/cidrs/aws/internal/ranges2go/genrawdata.sh
```
BenTheElder, line 99: this script changed in the last round of refactoring the cidr data.

Makefile
```diff
@@ -94,6 +94,10 @@ update:
 	hack/make-rules/update.sh
 tidy:
 	hack/make-rules/tidy.sh
+# generate ip ranges
+generate-ipranges:
```
BenTheElder, line 98: nit: IMHO this is a misleading name, I'd expect generate to do what the `go generate ./...` bit does (generate code from inputs), not to download new data.

maybe `update-ipranges`

udemyhoangit86 (approved):

PR #141, opened 2023-01-18
Handle unknown AWS regions explicitly

justinsb:
We were joining to an empty string, which is likely unintended.

### DIFF
--- a/cmd/archeio/app/buckets.go
+++ b/cmd/archeio/app/buckets.go
@@ -26,7 +26,9 @@ import (
 // awsRegionToS3URL returns the base S3 bucket URL for an OCI layer blob given the AWS region
 //
 // blobs in the buckets should be stored at /containers/images/sha256:$hash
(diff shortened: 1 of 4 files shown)


ameukam:
cc @BenTheElder

cmd/archeio/app/buckets.go
```diff
@@ -26,7 +26,9 @@ import (
 // awsRegionToS3URL returns the base S3 bucket URL for an OCI layer blob given the AWS region
 //
 // blobs in the buckets should be stored at /containers/images/sha256:$hash
-func awsRegionToS3URL(region string) string {
+//
+// If the region is not known, returns false.
+func awsRegionToS3URL(region string) (string, bool) {
```
BenTheElder, line 31: this method handles ALL possible regions as enforced by a unit test, so this makes the code more complicated for no gain?

BenTheElder:
this code got an overhaul and has a default bucket configured per instance instead

PR #146, opened 2023-02-07
Copy sync script over and use the bucket backing `gcr.io/k8s-artifacts-prod`

upodroid:
/cc @BenTheElder 

I copied over the script and changed the bucket name and script name.

/hold

I need an image to be pushed to `gcr.io/k8s-artifacts-prod` first and complete the initial sync at https://github.com/kubernetes/test-infra/pull/28652

### DIFF
--- a/hack/tools/sync-to-s3.sh
+++ b/hack/tools/sync-to-s3.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+# Copyright 2023 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


BenTheElder (commented):
/approve

I would review this more, but given it's just a copy and we know it works let's not bikeshed up front.

hack/tools/sync-to-s3.sh
```diff
@@ -0,0 +1,62 @@
...
+bucket_acl = private
+
+[s3]
+type = s3
+provider = AWS
+env_auth = true
+region = us-east-2
+EOF
+  echo "Running sync between '${SOURCE:-}' and '${DESTINATION:-}'"
+  if rclone sync --config "${RCLONE_CONFIG:-}" -P "${SOURCE:-}" "${DESTINATION:-}"; then
```
jhoblitt, line 59: Is this intended to be run periodically for awhile?  If so, it might be worth while to set an explicit timeout in the config and/or wrap the `rclone` invocation with `timeout` to prevent infinite hangs.

upodroid: Infinite hang shouldn't be a problem. Prow will kill the pod if it exceeds the max job duration and the build will fail and that should trigger notifications.

PR #151, opened 2023-02-10
MVP: "geranos" GCR/AR to S3 sync tool

BenTheElder:
Geranós is a tool to sync from GCR/AR to S3.

This is intended to eventually be general enough that it can be used to stand up a mirror similar to ours without depending on kubernetes specific details xref https://github.com/kubernetes/registry.k8s.io/issues/152, so we're listing the registry contents instead of reading promoter manifests or putting S3 in the image promoter. The promoter also has performance issues already, so instead this approach will replace the `rclone` script we have reading the exposed GCS bucket from GCR (see #146).

EDIT: To be clear: This is **NOT** ready to merge yet, I opened this PR just for visibility and to show that implementing this isn't too involved. We're going to need something like this to phase out the temporary GCR in https://github.com/kubernetes/enhancements/issues/3720 / https://github.com/kubernetes/enhancements/pull/3723#discussion_r1095205973

I still need to flesh out inputs / configuration, and testing. But this is more or less workable already.


cmd/geranos/README.md
```diff
@@ -0,0 +1,8 @@
+# Geranos
+
+γερανός (geranós) is Greek for "crane"
```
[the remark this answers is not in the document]
aojea: :)

cmd/geranos/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+This binary is a tool based on [crane] which is used to copy image layers
+from registries to object storage for backing [archeio](./../archeio)
+
+Currently it only supports Google Container Registry / Aritfact Registry to S3.
```
aojea, line 8:
```suggestion
Currently it only supports Google Container Registry / Artifact Registry to S3.
```

cmd/geranos/go.mod
```diff
@@ -0,0 +1,27 @@
+module k8s.io/registry.k8s.io/cmd/geranos
+
+go 1.19
+
+require (
+	github.com/google/go-containerregistry v0.12.0
+	k8s.io/klog v1.0.0
```
aojea, line 7: v1?

cmd/geranos/main.go
```diff
@@ -0,0 +1,145 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/google"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go/service/s3"
+	"github.com/aws/aws-sdk-go/service/s3/s3manager"
+
+	"k8s.io/klog"
```
aojea, line 33:
```suggestion
	"k8s.io/klog/v2"
```

cmd/geranos/main.go
```diff
@@ -0,0 +1,74 @@
...
+	// one of the backing registries for registry.k8s.io
+	// TODO: make configurable later
+	const sourceRegistry = "us-central1-docker.pkg.dev/k8s-artifacts-prod/images"
+
+	// TODO: make configurable later
+	const s3Bucket = "s3:prod-registry-k8s-io-us-east-2"
+
+	repo, err := name.NewRepository(sourceRegistry)
+	if err != nil {
+		klog.Fatal(err)
```
justinsb, line 37: Tip: I like `main()` to be `func main() { if err := run(); err != nil { fmt.Fprintf....` because then you don't have to rewrite from klog.Fatal -> fmt.Errorf

cmd/geranos/s3uploader.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (s *s3Uploader) CopyToS3(bucket string, layer v1.Layer) error {
+	key, err := keyForLayer(layer)
+	if err != nil {
+		return err
+	}
+	if s.skipExisting {
+		exists, err := s.blobExists(bucket, key)
+		if err != nil {
+			klog.Errorf("failed to check if blob exists: %v", err)
```
justinsb, line 56: Nit: probably should not upload here?  Or if there's a common error case here, a comment would be helpful!

cmd/geranos/s3uploader.go
```diff
@@ -0,0 +1,97 @@
...
+			return nil
+		}
+	}
+	r, err := layer.Compressed()
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+	_, err = s.uploader.Upload(&s3manager.UploadInput{
+		Bucket: aws.String(bucket),
```
justinsb, line 67: Nit: Should probably check that this is doing a checksum-ed upload.  If not we might need to copy to a temp file before uploading...

cmd/geranos/walkmanifests.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/google"
+)
+
+type walkManifestsFunc func(ref name.Reference) error
+
+// Unfortunately this is only doable on GCP currently.
+//
+// TODO: To support other registries in the meantime, we could require a list of
```
justinsb, line 30: You could use the promotion manifests, I guess

go.work
```diff
@@ -0,0 +1,7 @@
+go 1.19
```
justinsb, line 1: Supposedly you don't want to check this in.  I'd suggest avoiding multiple modules if you can, but /shrug

justinsb (commented):
lgtm

cmd/geranos/s3uploader.go
BenTheElder: It's unclear what an expected error would be here (bad API call??), but uploading again should be a harmless (just costs more time) fall through

cmd/geranos/main.go
BenTheElder: this is very MVP still :-)

cmd/archeio has better structure, similar to what you're suggesting.

go.work
BenTheElder: isolating multiple modules used to be more important for keeping dependency trees manageable (see also tools module pattern). I think that's no longer necessary, tidy seems to be getting smarter.

BenTheElder: I started working on this a while back, this is one of the errant bits I should clean up before merge.

The PR _is_ titled WIP and the description mentions it's just for visibility, and not ready yet :-)

cmd/geranos/s3uploader.go
BenTheElder: I think it does, but the upload code is also a bit of a stub, I haven't wired in auth / actually uploaded anything anywhere yet.

cmd/geranos/main.go
BenTheElder: in the first commit, this method contained essentially all logic, in a single file main.go as a proof of concept.

i've been teasing out bits, it's pretty small now but not there yet.

go.work
BenTheElder: Seems go1.17 is when this got fixed, and only for modules that declare themselves as 1.17 or higher https://go.dev/ref/mod#graph-pruning

cmd/geranos/walkmanifests.go
BenTheElder: For Kubernetes I'm not worried about this.

I'd like the tool to be reusable by other projects, and currently that requires the use of GCR/AR (or using their own tool to populate the blob storage while reusing archeio), but that is low-priority to fix versus getting ourselves off of rclone.

go.work
BenTheElder: keeping the tools module for now because the dep graph is messier and _only_ used at test time (and we still have to have no conflicts between those version and the libraries we _actually_ import if we drop that module), but dropped the geranos module and go.work

cmd/geranos/s3uploader.go
BenTheElder: Or reworded: I think failing towards "just upload it anyhow" is reasonable because skipExisting is only intended to speed up synchronizing. Failing to ensure all content is uploaded to mirrors can cost us $$$ in bandwidth.

I don't think we reasonably expect an error here, but instead of just treating errors as "it doesn't exist" we at least also log them.

cmd/geranos/walkimages.go
```diff
@@ -0,0 +1,87 @@
...
+
+	// google.Walk already resolves these to individual manifests
+	if desc.MediaType.IsIndex() {
+		return nil
+	}
+
+	// It's unlikely we need to support these
+	// https://github.com/google/go-containerregistry/issues/377
+	if desc.MediaType == types.DockerManifestSchema1 || desc.MediaType == types.DockerManifestSchema1Signed {
+		klog.Warning("Skipping ancient schema 1 image: " + ref.String())
```
BenTheElder, line 72: I ran a full scan, we have 3597 of these.

We however should not expect any _new_ images of this type, and we already have the rclone backfill.

BenTheElder (commented):
Current unoptimized state takes just under 11 minutes to walk everything (no actual content uploads). Not ideal but not super concerning, given we just want to keep after new images automatically, even if not the moment they're first published. 

We'll definitely want to at least parallelize the actual content upload though.

First log: 19:33:54
Last Log: 01:44:32

Outmoded Schema V1 / Schema V1 Signed hits: 3597
Other Images: 82503

Skimming those, as expected they're generally old tags for which newer tags are in contemporary format.

If we _really_ need to we can add Schema V1 later with a bit of refactoring, not going to focus on that for now.

cmd/geranos/main.go
BenTheElder: this is done, closer to app/archeio. will do even more refactoring later though, I want to keep it smaller for the MVP just to show that implementing this quickly is doable, before we add the other boilerplate like input options

cmd/geranos/walkimages.go
BenTheElder: Schema v1 is now implemented.

cmd/geranos/s3uploader.go
BenTheElder: Explicitly setting sha256 checksum from the layer digest now.

BenTheElder:
~22 minutes to scan 86,116 images and dry-run upload (everything up to actually copying to S3) 439,787 layers on my laptop. Copy time may still be an issue, but quick-syncs skipping existing layers would probably have a similar runtime to this dry-run.

Now supports schema v1 images.

Still needs cleanup, configurability, tests.

aojea:
lgtm

BenTheElder:
I'm going to say let's go ahead with this and iterate, to keep the PR sizes reasonable.

Next PR I'll wire up inputs for a few of the fixed things, then we can get an optional periodic running and see if there's any issues with the S3 configuration.

Then I'll work on some more optimizations before we switch over.

cmd/geranos/walkimages.go
```diff
@@ -0,0 +1,130 @@
...
+// TODO: To support other registries in the meantime, we could require a list of
+// image names as an input and plumb that through, then list tags and get something
+// close to this. The _catalog endpoint + tag listing could also work in some cases.
+//
+// However, even then, this is more complete because it lists all manifests, not just tags.
+// It's also simpler and more efficient.
+//
+// See: https://github.com/opencontainers/distribution-spec/issues/222
+func WalkImageLayersGCP(repo name.Repository, walkImageLayers WalkImageLayersFunc) error {
+	g := new(errgroup.Group)
```
justinsb, line 48: TIL

justinsb (approved):

justinsb:
Additive, so low risk => merge and iterate SGTM

/approve
/lgtm

Issue #153, opened 2023-02-14
carefully consider enabling artifact registry user quotas


BenTheElder:
Currently quotas are per region/project. For us that means each AR region.

https://cloud.google.com/artifact-registry/quotas#project-quota

> 60000 requests per minute in each region or multi-region.
> 18000 write or delete requests per minute in each region or multi-region.
>
> In most cases, a single HTTP request or API call counts as a single request. However, some operations count as multiple requests. For example, a batch request like ImportAptArtifacts might charge quota for each item in the batch. A Docker pull or push usually makes multiple HTTP requests, so quota is charged for each request.

We've seen https://kubernetes-sigs/promo-tools and now #151 while in development (excessive concurrency) hit `429 Too Many Requests`, which currently causes an outage for the involved region during that 1-minute quota window.

We should fix our tools (I've throttled #151) but intentional abuse is also a concern.

> Per-user request quota
> By default, projects have unlimited per-user quota. You can optionally [cap](https://cloud.google.com/docs/quota#capping_usage) per-user quota within a project. Per-user quota applies per authenticated user or per client IP address for unauthenticated requests to a public repository.

We should look at enabling this. It's worth noting that doing so may break the image promoter if it exceeds the new per-user quota, but we can adjust for that.

We can also request a quota increase. Jon suggested 2x for the busier regions would probably be reasonable for us.

BenTheElder:
@ameukam filed an initial quota increase request.

We still need to make sure that we can safely enable per-user quotas (we're a bit of an odd user), including considering quota needs for promo-tools.

Setting a per-user quota should at least help us avoid accidentally DOSing ourselves.
We may need to consider further action to deal with malicious intent.

ameukam:
Filled 2x of default quota (60000) for:
europe-west2
europe-west4
us-central1
us-east4

ameukam:
Request have been approved for only 2 regions:

<img width="1171" alt="image" src="https://user-images.githubusercontent.com/2343515/219301168-07f6352f-c034-45c4-9f1f-ee2a70db2606.png">

BenTheElder:
I can't create public read AR instances at work, but we could enable it on one of the low-traffic regions and run some tests before rolling it out to more regions.

GCR had:

> There is a fixed quota limit on requests to Container Registry hosts. This limit is per client IP address.
> 
> 50,000 HTTP requests every 10 minutes
> 1,000,000 HTTP requests per day

[...]

PR #155, opened 2023-02-17
iterate on geranos MVP

BenTheElder:
couple quick follow-ups that should be enough to get a test job running:

make dry-run configurable
throttle concurrency more ref #153 
fix s3 bucket / keys
skip existing layers by default

### DIFF
--- a/cmd/geranos/main.go
+++ b/cmd/geranos/main.go
@@ -43,20 +43,16 @@ func Run(_ []string) error {
 	const sourceRegistry = "us-central1-docker.pkg.dev/k8s-artifacts-prod/images"
 
 	// TODO: make configurable later
-	const s3Bucket = "s3:prod-registry-k8s-io-us-east-2"
+	const s3Bucket = "prod-registry-k8s-io-us-east-2"
 
 	repo, err := name.NewRepository(sourceRegistry)
 	if err != nil {
 		return err
 	}
-	s3Uploader, err := newS3Uploader()
+	s3Uploader, err := newS3Uploader(os.Getenv("REALLY_UPLOAD") == "")
 	if err != nil {
 		return err
 	}
-
-	// for debugging: force this true
-	// TODO: make configurable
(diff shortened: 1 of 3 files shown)


cmd/geranos/main.go
```diff
@@ -43,21 +43,17 @@ func Run(_ []string) error {
...
 	// TODO: make configurable later
-	const s3Bucket = "s3:prod-registry-k8s-io-us-east-2"
+	const s3Bucket = "prod-registry-k8s-io-us-east-2"
 
 	repo, err := name.NewRepository(sourceRegistry)
 	if err != nil {
 		return err
 	}
-	s3Uploader, err := newS3Uploader()
+	s3Uploader, err := newS3Uploader(os.Getenv("REALLY_UPLOAD") == "")
```
TerryHowe, line 52:
```suggestion
	s3Uploader, err := newS3Uploader(os.Getenv("DRY_RUN") != "")
```
I'm new here, so maybe this makes more sense to everyone else.

cmd/geranos/s3uploader.go
```diff
@@ -32,22 +33,32 @@ import (
...
+const blobKeyPrefix = "containers/images/"
 
 type s3Uploader struct {
-	svc          *s3.S3
-	uploader     *s3manager.Uploader
-	skipExisting bool
-	dryRun       bool
+	svc            *s3.S3
+	uploader       *s3manager.Uploader
+	reuploadLayers bool
```
TerryHowe, line 42: Doesn't look like this is used

cmd/geranos/main.go
BenTheElder: The default for now is to do a dry run. This would invert that default.

cmd/geranos/s3uploader.go
BenTheElder: It is read. It is not yet written.

justinsb:
LGTM

/lgtm

PR #161, opened 2023-03-03
don't upload config layers for now

BenTheElder:
to match the rclone job, and focus on replacing it.


ameukam:
/lgtm
/hold

Remove the hold when ready.

PR #163, opened 2023-03-08
reduce geranos concurrency

BenTheElder:
xref #153 

this is a quick patch, it's not the end-game.


BenTheElder:
we can avoid a lot of API calls by keeping an index, but speed is not maximally urgent here and this is the smallest diff. I think this value will be sufficient but we can iterate.

BenTheElder:
triggered a rerun https://prow.k8s.io/?job=*s3*

PR #164, opened 2023-03-08
reject requests for /v2/_catalog

BenTheElder:
cc @jonjohnsonjr

pros:
- _catalog is non-portable, not supporting it leaves us more flexibility in what registry we depend on
- _catalog is scoped to a registry host, not a repository

cons:
- _catalog is useful, implementing it would be more helpful (but also then we need to continue to support it...)

I think blocking the broken call (we redirect to a path prefixed _catalog which won't work anywhere) is better than leaving a broken one. We can continue discussing potentially implementing in the meantime.

### DIFF
--- a/cmd/archeio/app/handlers.go
+++ b/cmd/archeio/app/handlers.go
@@ -99,6 +99,12 @@ func makeV2Handler(rc RegistryConfig, blobs blobChecker) func(w http.ResponseWri
 			w.WriteHeader(http.StatusOK)
 			return
 		}
+		// we don't support the non-standard _catalog API
+		// https://github.com/kubernetes/registry.k8s.io/issues/162
+		if rPath == "/v2/_catalog" {
+			http.Error(w, "_catalog is not supported", http.StatusNotFound)
+			return
+		}
 
 		// check if blob request
 		matches := reBlob.FindStringSubmatch(rPath)
--- a/cmd/archeio/app/handlers_test.go
+++ b/cmd/archeio/app/handlers_test.go
@@ -186,6 +186,15 @@ func TestMakeV2Handler(t *testing.T) {
 			ExpectedStatus: http.StatusBadRequest,
 		},
 		{
+			Name: "/v2/_catalog",
+			Request: func() *http.Request {
+				r := httptest.NewRequest("GET", "http://localhost:8080/v2/_catalog", nil)
+				r.RemoteAddr = "35.180.1.1:888"
+				return r
+			}(),
+			ExpectedStatus: http.StatusNotFound,
+		},
+		{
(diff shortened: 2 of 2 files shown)


BenTheElder:
fixes #164

cmd/archeio/app/handlers.go
```diff
@@ -99,6 +99,12 @@ func makeV2Handler(rc RegistryConfig, blobs blobChecker) func(w http.ResponseWri
 			w.WriteHeader(http.StatusOK)
 			return
 		}
+		// we don't support the non-standard _catalog API
+		// https://github.com/kubernetes/registry.k8s.io/issues/162
+		if rPath == "/v2/_catalog" {
```
BenTheElder, line 104: should probably add a test-case for this.

cmd/archeio/app/handlers.go
```diff
@@ -99,6 +99,12 @@ func makeV2Handler(rc RegistryConfig, blobs blobChecker) func(w http.ResponseWri
...
+			http.Error(w, "_catalog is not supported", http.StatusNotFound)
```
TerryHowe, line 105:
```suggestion
			http.Error(w, "_catalog is not supported see <a href='https://explore.ggcr.dev/?repo=registry.k8s.io'>https://explore.ggcr.dev/?repo=registry.k8s.io</a>", http.StatusNotFound)
```

TerryHowe: Not sure if that adds much value though because not sure people would see it

BenTheElder: I don't think we should serve HTML here, the API under `/v2` is not normally viewed in a browser.

BenTheElder: (actually the spec expects JSON statuses, but real production registries do serve simple human readable text with their error codes)

ameukam:
/lgtm
/hold

Remove the hold when ready

BenTheElder:
Added test case covering this change.

We're missing a little test coverage in archeio from other changes, regressed to:

> ok  	k8s.io/registry.k8s.io/cmd/archeio/app	1.590s	coverage: 95.7% of statements

Will fix that in another PR since unrelated.

cmd/archeio/app/handlers.go
BenTheElder: done.

ameukam:
As expected:

```bash
http -I https://registry-sandbox.k8s.io/v2/_catalog                                                                                                5s  07:22:34 AM
HTTP/1.1 404 Not Found
Alt-Svc: h3=":443"; ma=2592000,h3-29=":443"; ma=2592000
Content-Length: 26
Date: Thu, 09 Mar 2023 06:32:17 GMT
Server: Google Frontend
Via: 1.1 google
X-Cloud-Trace-Context: 230d24da1c3b4d08a2e6301497da2db5;o=1
content-type: text/plain; charset=utf-8
x-content-type-options: nosniff

_catalog is not supported
```

PR #165, opened 2023-03-09
rate limit most AR calls

BenTheElder:
- block on rate limiter before getting image manifest
- block on rate limiter before getting image layer
- rate limit to 40 QPS AKA 2400 RPM

ref: #153


BenTheElder:
next run isn't done yet but good so far 6 minutes in, i think we're good again
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ar-to-s3-sync/1633655548832387072

PR #166, opened 2023-03-09
rate limit all AR API calls

BenTheElder:
more complete and invasive version of #165 

worth noting: 

> In most cases, a single HTTP request or API call counts as a single request. However, some operations count as multiple requests. For example, a batch request like ImportAptArtifacts might charge quota for each item in the batch. A Docker pull or push usually makes multiple HTTP requests, so quota is charged for each request.

https://cloud.google.com/artifact-registry/quotas

### DIFF
--- a/cmd/geranos/main.go
+++ b/cmd/geranos/main.go
@@ -45,6 +45,10 @@ func Run(_ []string) error {
 	// TODO: make configurable later
 	const s3Bucket = "prod-registry-k8s-io-us-east-2"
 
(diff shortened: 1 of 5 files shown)


BenTheElder:
running in https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ar-to-s3-sync/1633710630219288576

BenTheElder:
worked, but slow. we can increase the client-side rate limit a bit, and look at shipping the optimizations to skip already synced images more.

xref: #153

PR #170, opened 2023-03-09
increase geranos registry rate limit

BenTheElder:
ref #166 #153 

This is working fine now, but a bit slower than desirable.

I'm going to work on other improvements, but in the meantime we might as well push closer to the rate limit and reduce the sync time hit (~13m => ~1h30m).


BenTheElder:
https://prow.k8s.io/?job=*s3*

drove runs down from ~1h31m to ~1h8m

PR #173, opened 2023-03-11
archeio: support blob digest algorithms other than sha256

BenTheElder:
A bit of future-proofing. 

As-is we'd have just redirected to the upstream registry for all non-sha256 blobs, but after this change we should still redirect to S3 if the blob is available.

geranos should already be able to upload arbitrary digest keys, though it can only instruct S3 to integrity check uploads with sha256 currently.

### DIFF
--- a/cmd/archeio/app/handlers.go
+++ b/cmd/archeio/app/handlers.go
@@ -72,7 +72,12 @@ func MakeHandler(rc RegistryConfig) http.Handler {
 
 func makeV2Handler(rc RegistryConfig, blobs blobChecker) func(w http.ResponseWriter, r *http.Request) {
 	// matches blob requests, captures the requested blob hash
-	reBlob := regexp.MustCompile("^/v2/.*/blobs/sha256:([0-9a-f]{64})$")
+	// https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pull
+	// Blobs are at `/v2/<name>/blobs/<digest>`
+	// Note that ':' cannot be contained in <name> but *must* be contained in <digest>
+	// <digest> also cannot contain `/` so we can use a relatively simple and cheap regex
+	// to match blob requests and capture the digest
+	reBlob := regexp.MustCompile("^/v2/.*/blobs/([^/]+:[a-zA-Z0-9=_-]+)$")
 	// initialize map of clientIP to AWS region
 	regionMapper := cloudcidrs.NewIPMapper()
 	// capture these in a http handler lambda
@@ -116,7 +121,7 @@ func makeV2Handler(rc RegistryConfig, blobs blobChecker) func(w http.ResponseWri
 			return
 		}
 		// it is a blob request, grab the hash for later
-		hash := matches[1]
+		digest := matches[1]
 
 		// for blob requests, check the client IP and determine the best backend
 		clientIP, err := clientip.Get(r)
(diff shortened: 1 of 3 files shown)


cmd/archeio/app/handlers_test.go
```diff
@@ -160,15 +160,15 @@ func TestMakeV2Handler(t *testing.T) {
...
-			"https://prod-registry-k8s-io-eu-west-1.s3.dualstack.eu-west-1.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
-			"https://prod-registry-k8s-io-us-east-1.s3.dualstack.us-east-2.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
-			"https://prod-registry-k8s-io-us-east-2.s3.dualstack.us-east-2.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
-			"https://prod-registry-k8s-io-us-west-1.s3.dualstack.us-west-1.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
-			"https://prod-registry-k8s-io-us-west-2.s3.dualstack.us-west-2.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
-			"https://prod-registry-k8s-io-eu-west-2.s3.dualstack.eu-west-2.amazonaws.com/containers/images/sha256%3Ada86e6ba6ca197bf6bc5e9d900febd906b133eaa4750e6bed647b0fbe50ed43e":           true,
```
BenTheElder, line 171: percent encoding these was unnecessary. both forms work. not percent encoding is simpler.

cmd/archeio/app/handlers.go
```diff
@@ -72,7 +72,12 @@ func MakeHandler(rc RegistryConfig) http.Handler {
 
 func makeV2Handler(rc RegistryConfig, blobs blobChecker) func(w http.ResponseWriter, r *http.Request) {
 	// matches blob requests, captures the requested blob hash
-	reBlob := regexp.MustCompile("^/v2/.*/blobs/sha256:([0-9a-f]{64})$")
+	// https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pull
+	// Blobs are at `/v2/<name>/blobs/<digest>`
+	// Note that ':' cannot be contained in <name> but *must* be contained in <digest>
+	// <digest> also cannot contain `/` so we can use a relatively simple and cheap regex
+	// to match blob requests and capture the digest
+	reBlob := regexp.MustCompile("^/v2/.*/blobs/([^/]+:[a-zA-Z0-9=_-]+)$")
```
dims, line 80: @BenTheElder this regexp compile happens only once right?

cmd/archeio/app/handlers_test.go
dims: ack

cmd/archeio/app/handlers.go
BenTheElder: Yes, and we're only swapping the existing regex for one that permits and captures arbitrary digest algorithms.

BenTheElder:
/hold cancel

Confident in this one but we'll also soak in staging.

I did some additional testing besides our integration (actually pulls an image through a local instance) and unit tests.

This is unlikely to be necessary at the moment but I could see not doing it biting us later, want to avoid a situation where down the line all traffic suddenly reverts to the source registry. We'd at least be online then but our costs would spike badly ...

PR #147, opened 2023-02-08
Default to AWS when possible

BenTheElder:
Implements #143 

TODO: Configure default s3 backend per cloud run region in the production config via env before we ship this. Possibly use cloudfront for that purpose instead in the future.
EDIT: see https://github.com/kubernetes/k8s.io/pull/4739

This is a messy change, a lot of things needed renaming / refactoring. 
The pre-parsed generator is still a bit gross, but fairly well tested and covers GCP + AWS now.

PR #175, opened 2023-03-13
clarify archeio docs

BenTheElder:
- We should discuss the current design before the original design
- Clarifying that you *must* join the mailinglist to access the original design sketch Google doc, still getting a lot of requests for this doc and this is the only place we currently link it AFAIK. Also clarifying that it's *not* current and the design has evolved, and outlining how that has evolved a bit.
- Adding another pointer to the main readme and particularly the stability section
- Clarifying repeatedly that while we document for transparency, you cannot rely on the implementation details
- Adding a bit more context re: security posture

I need to do a follow-up updating the "request handling" diagrams, we've made some slight changes since the last update.

### DIFF
--- a/cmd/archeio/README.md
+++ b/cmd/archeio/README.md
@@ -7,14 +7,6 @@ OCI artifact ("docker image") hosting.
 
 Current design details will be detailed here as they mature.
 
-The original design doc is shared with members of
-[dev@kubernetes.io](https://groups.google.com/a/kubernetes.io/g/dev), 
-anyone can join this list and gain access to read
-[the document](https://docs.google.com/document/d/1yNQ7DaDE5LbDJf9ku82YtlKZK0tcg5Wpk9L72-x2S2k/). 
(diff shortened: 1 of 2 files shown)


BenTheElder:
I think actually the only thing missing from request handling doc is `/v2/_catalog`, we did update it when doing #147

cmd/archeio/README.md
```diff
@@ -31,20 +25,60 @@ For a rough TLDR of the current design:
 - Mirrors *of content* blobs are hosted in S3 buckets in AWS
 - AWS clients are detected by client IP address and redirect to a local S3 bucket copy
 *only* when requesting content blobs, *not* manifests, manifest lists, tags etc.
-- All other requests are redirected to the original upstream registry
+- GCP clients and other requests are rerouted to the regional upstream source ARtifact Registries
```
dims, line 28: lowercase

PR #176, opened 2023-03-14
Update ko version

TerryHowe:


hack/tools/go.mod
```diff
@@ -4,25 +4,24 @@ go 1.19
 
 require (
 	github.com/golangci/golangci-lint v1.51.1
-	github.com/google/go-containerregistry v0.13.0
-	// TODO: switch back to next stable release
-	github.com/google/ko v0.12.1-0.20230206135429-568a51e6bcbc
+	github.com/google/go-containerregistry v0.13.1-0.20230310164735-e94d40893b2d
```
BenTheElder, line 7: Is the ko update forcing us off a tagged container registry?

hack/tools/go.mod
```diff
@@ -4,25 +4,24 @@ go 1.19
...
+	github.com/google/ko v0.13.0
```
BenTheElder, line 8: Any notable changes we should be aware of? Did we check for any action required?

TerryHowe: I'm not sure why this TODO was created, so I don't know which fix was needed.\ I don't see any breaking changes https://github.com/ko-build/ko/releases/tag/v0.13.0

Support for the idea: https://github.com/kubernetes/release/pull/2961

All the tests are working, but admittedly I don't have a great way to test and validate this.

BenTheElder: The TODO is because kubernetes prefers to stick to tagged releases of dependencies, but we temporarily upgraded to a version between releases pickup an essential unreleased fix.

we'd generally check the release notes, just haven't gotten to that yet myself and asking if you have 👀

BenTheElder: Extra sensitive to any breaking changes for the next month at least while we try to ship the k8s.gcr.io redirect, want to be more thorough than the tests even :-)

TerryHowe: Another day then

BenTheElder: I read through this and it looks safe? We just need to be thorough 😅

BenTheElder (commented):
/lgtm
/approve

PR #178, opened 2023-03-15
fix dependabot alerts

BenTheElder:
nothing in registry.k8s.io production deployment (geranos, tools, not archeio), but still, better to clear these so we can continue to monitor open alerts


go.mod
```diff
@@ -25,9 +25,9 @@ require (
 	github.com/opencontainers/image-spec v1.1.0-rc2 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
 	github.com/sirupsen/logrus v1.9.0 // indirect
-	golang.org/x/net v0.1.0 // indirect
+	golang.org/x/net v0.8.0 // indirect
 	golang.org/x/oauth2 v0.1.0 // indirect
-	golang.org/x/sys v0.1.0 // indirect
+	golang.org/x/sys v0.6.0 // indirect
```
BenTheElder, line 30: auto-bumped when bumping x/net.

same thing, only used in geranos currently.

BenTheElder:
back to no open alerts now 👍

PR #179, opened 2023-03-15
enforce code coverage levels

BenTheElder:
Quick starting point to fix a long-standing TODO

I couldn't find a great off the shelf tool to simply enforce coverage levels from cover files. There are one or two tools but they tend to require also running the tests.

Go makes it very simple to parse coverage output, so I just wrote a tiny utility to require 100% coverage with some excluded packages. Primarily this ensures production archeio code maintains current coverage levels.

### DIFF
--- a/cmd/archeio/docs/testing.md
+++ b/cmd/archeio/docs/testing.md
@@ -12,7 +12,7 @@ methods, we also have unit tests covering the HTTP Handlers and full
 
 **This level of coverage must be maintained**, it is imperative that we have robust
 testing in this project that may soon serve all Kubernetes project image downloads.
-TODO: this should be enforced by CI. Currently it is enforced by reviewers.
+We automatically enforce 100% code coverage for archeio sources.
 
 Coverage results are visible by clicking the `pull-oci-proxy-test` context link
 at the bottom of the pull request.
--- a/hack/tools/require-coverage/main.go
+++ b/hack/tools/require-coverage/main.go
@@ -59,7 +59,7 @@ var knownFailingFiles = sets.NewString(
 
 func main() {
 	fmt.Println("Checking coverage ...")
-	profiles, err := cover.ParseProfiles("./../../bin/all-filtered.cov")
+	profiles, err := cover.ParseProfiles("./../../bin/all.cov")
 	if err != nil {
 		panic(err)
 	}
--- a/hack/tools/require-coverage/main.go
+++ b/hack/tools/require-coverage/main.go
@@ -39,8 +39,6 @@ var knownFailingFiles = sets.NewString(
 	// this code is used only at development time and integration testing it
 	// is probably excessive
 	"k8s.io/registry.k8s.io/pkg/net/cloudcidrs/internal/ranges2go/main.go",
-	// TODO: this is reasonable to test and has poor coverage currently
-	"k8s.io/registry.k8s.io/pkg/net/cloudcidrs/internal/ranges2go/parse_gcp.go",
 	// TODO: this is reasonable to test but shy of 100% coverage, mostly error handling ...
 	"k8s.io/registry.k8s.io/pkg/net/cloudcidrs/internal/ranges2go/gen.go",
 	// geranos is not easily tested and is not in the blocking path in production
--- a/hack/tools/require-coverage/main.go
+++ b/hack/tools/require-coverage/main.go
@@ -1,0 +1,99 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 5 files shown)


hack/tools/require-coverage/main.go
```diff
@@ -0,0 +1,99 @@
...
+		fmt.Println("All code coverage either acceptable or ignored")
+	}
+}
+
+func coverPercent(profile *cover.Profile) float64 {
+	totalStatements := 0
+	coveredStatements := 0
+	for _, block := range profile.Blocks {
+		totalStatements += block.NumStmt
+		if block.Count > 0 {
```
BenTheElder, line 94: count is how many times the block executed. 
it might be >1 if we're in count mode, or in other modes it might only be 1 or zero.

this is why we sum the statements covered in each non-zero block

cmd/archeio/docs/testing.md
```diff
@@ -12,7 +12,7 @@ methods, we also have unit tests covering the HTTP Handlers and full
 
 **This level of coverage must be maintained**, it is imperative that we have robust
 testing in this project that may soon serve all Kubernetes project image downloads.
-TODO: this should be enforced by CI. Currently it is enforced by reviewers.
+We automatically enforce 100% code coverage for archeio sources.
```
aojea, line 15: is this for real?

BenTheElder: yes*, after this PR :-)

test test presubmit runs this tool now https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/registry.k8s.io/179/pull-oci-proxy-test/1635909009225551872

\* not on main.go ... yet. I need to rework the integration tests. that's another PR, prioritizing this one. also not zz_generated* which is filtered from all.cov => filtered.cov, but I think we can remove filtering that and we definitely cover it actually.

we're excluding other sources currently, such as the cmd/geranos sources, and some of the ip ranges code generator

BenTheElder: ... and I just added a commit to stop filtering out generated files. so just cmd/archeio/main.go needs fixing.

BenTheElder:
```
Checking coverage ...
PASSED: k8s.io/registry.k8s.io/cmd/archeio/app/buckets.go 100%
PASSED: k8s.io/registry.k8s.io/cmd/archeio/app/handlers.go 100%
IGNORE: k8s.io/registry.k8s.io/cmd/archeio/main.go 0%
IGNORE: k8s.io/registry.k8s.io/cmd/geranos/main.go 0%
IGNORE: k8s.io/registry.k8s.io/cmd/geranos/ratelimitroundtrip.go 0%
IGNORE: k8s.io/registry.k8s.io/cmd/geranos/s3uploader.go 0%
IGNORE: k8s.io/registry.k8s.io/cmd/geranos/schemav1.go 0%
IGNORE: k8s.io/registry.k8s.io/cmd/geranos/walkimages.go 0%
IGNORE: k8s.io/registry.k8s.io/internal/integration/bins.go 75%
IGNORE: k8s.io/registry.k8s.io/internal/integration/paths.go 81.25%
PASSED: k8s.io/registry.k8s.io/pkg/clientip/clientip.go 100%
PASSED: k8s.io/registry.k8s.io/pkg/net/cidrs/bruteforce.go 100%
PASSED: k8s.io/registry.k8s.io/pkg/net/cidrs/triemap.go 100%
IGNORE: k8s.io/registry.k8s.io/pkg/net/cloudcidrs/internal/ranges2go/gen.go 76.92307692307693%
IGNORE: k8s.io/registry.k8s.io/pkg/net/cloudcidrs/internal/ranges2go/main.go 0%
PASSED: k8s.io/registry.k8s.io/pkg/net/cloudcidrs/internal/ranges2go/parse_aws.go 100%
IGNORE: k8s.io/registry.k8s.io/pkg/net/cloudcidrs/internal/ranges2go/parse_gcp.go 65.51724137931035%
PASSED: k8s.io/registry.k8s.io/pkg/net/cloudcidrs/internal/ranges2go/prefix_helpers.go 100%
PASSED: k8s.io/registry.k8s.io/pkg/net/cloudcidrs/ipmapper.go 100%
All code coverage either acceptable or ignored
```

cmd/archeio/docs/testing.md
aojea: I always found 100% excessive, my observation is that around 80 or 85 is a good coverage level

aojea:
impressing

pkg/net/cloudcidrs/internal/ranges2go/gen_test.go
```diff
@@ -23,7 +23,7 @@ import (
 
 func TestGenerateRangesGo(t *testing.T) {
```
BenTheElder, line 24: small test case for next commit

cmd/archeio/docs/testing.md
BenTheElder: That might generally be true but I don't think so for a small stateless HTTP redirector.

we can and absolutely should test all code paths. There's a high risk to production workloads if we have any bugs and the service logic has minimal external dependencies that are easily mocked (HEAD requests to object storage)

some of the coverage is more integration test than unit but go doesn't distinguish and  happy to count both as coverage

BenTheElder: We're not requiring any particular level currently for the "ignored" files which are used for other less critical tools.

it's also ironically harder to effectively e2e test all codepaths in the registry server than unit / integration because we route based on client address so ...

PR #188, opened 2023-03-21
update README links

BenTheElder:
We've moved k8s.gcr.io/ => registry.k8s.io/ for publishing in k8s.io repo

Also cleaning up the space spent on URLs vs hyperlinking text in the precious section at the top where we need to avoid losing people's attention before the stability section.


BenTheElder:
superseded by https://github.com/kubernetes/registry.k8s.io/pull/189

PR #200, opened 2023-03-23
use community images to list images to mirror

BenTheElder:
following https://github.com/kubernetes-sigs/community-images/issues/47 we can point to a general tool to dump the running images in a cluster*

*excluding at least pause. getting that info from a running cluster portably would be quite a trick.

### DIFF
--- a/docs/mirroring/README.md
+++ b/docs/mirroring/README.md
@@ -29,7 +29,10 @@ help you identify Kubernetes Project image references to mirror like this:
 kubectl community-images --mirror
 ```
 
-We also have tooling specific guides:
+**NOTE**: This will only find images specified in your currently running pods,
+and not for example the "pause" image used to implement pods in containerd / cri-o / dockershim.
+
+Tooling specific guides:
 
 - For containerd see: [containerd.md](./containerd.md)
 - For kubeadm see: [kubeadm.md](./kubeadm.md)
--- a/docs/mirroring/README.md
+++ b/docs/mirroring/README.md
@@ -22,6 +22,15 @@ See: https://kubernetes.io/docs/contribute/style/content-guide/#dual-sourced-con
 
 <!--TODO: Generically identifying registry.k8s.io images in manifests / charts / addons.-->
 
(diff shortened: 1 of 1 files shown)


docs/mirroring/README.md
```diff
@@ -22,6 +22,15 @@ See: https://kubernetes.io/docs/contribute/style/content-guide/#dual-sourced-con
 
 <!--TODO: Generically identifying registry.k8s.io images in manifests / charts / addons.-->
 
+If you have a running cluster then our [community-images] [krew] plugin can
+help you identify Kubernetes Project image references to mirror like this:
+
+```console
+kubectl community-images --mirror
```
BenTheElder, line 29: putting this tool inline because it's very very broadly useful to inventory images without much context on what the user is using. we should expand on this a bit.

docs/mirroring/README.md
```diff
@@ -22,6 +22,15 @@ See: https://kubernetes.io/docs/contribute/style/content-guide/#dual-sourced-con
 
 <!--TODO: Generically identifying registry.k8s.io images in manifests / charts / addons.-->
 
+If you have a running cluster then our [community-images] [krew] plugin can
```
BenTheElder, line 25: [krew] missing link

BenTheElder, line 25: having two links side by side is a little hard to distinguish, maybe re-wording to space them slightly

BenTheElder: or we might not need a krew link anyhow since community-images covers that?

BenTheElder: dropped the krew link

BenTheElder:
waiting for https://github.com/kubernetes-sigs/community-images/issues/56

then will push the shell snippet here and cleanup

BenTheElder:
plugin is fixed
/hold cancel

docs/mirroring/README.md
BenTheElder: expanded a bit more in later commit

PR #201, opened 2023-03-24
expand kOps and containerd mirroring guides

BenTheElder:
trying to chip away at these in-between rollouts.

### DIFF
--- a/docs/mirroring/README.md
+++ b/docs/mirroring/README.md
@@ -32,10 +32,11 @@ kubectl community-images --mirror
 **NOTE**: This will only find images specified in your currently running pods,
 and not for example the "pause" image used to implement pods in containerd / cri-o / dockershim.
 
-Tooling specific guides:
+For specific tools we have these guides:
 
 - For containerd see: [containerd.md](./containerd.md)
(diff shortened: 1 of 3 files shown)


docs/mirroring/README.md
```diff
@@ -109,7 +110,13 @@ Instead, link to existing official guides and merely provide a lightweight point
 See: https://kubernetes.io/docs/contribute/style/content-guide/#dual-sourced-content
 -->
 
-<!--TODO: cri-o, general manifests-->
+<!--TODO: cri-o-->
+
+A simple option for many cases is to update the `image` fields in your
```
BenTheElder, line 115: self-nit post-merge: I think we stopped using "simple" in docs because it assumes level of expertise etc.

will clean that up in a follow-up

BenTheElder: Yep: https://kubernetes.io/docs/contribute/style/style-guide/#avoid-words-that-assume-a-specific-level-of-understanding

BenTheElder: https://github.com/kubernetes/registry.k8s.io/pull/202

PR #205, opened 2023-03-24
add links to skip ahead below each mirroring option

BenTheElder:
this will be particularly useful when we have more options


BenTheElder:
/hold
I think I broke these

BenTheElder:
/hold cancel
only in the rich diff viewer, they work fine viewing the file

TerryHowe (approved):
/lgtm

Issue #181, opened 2023-03-16
🚨 Registry Code Freeze 🚨


BenTheElder:
We will only accept critical bug fixes to `cmd/archeio` and related code for a while until this tracking issue is closed.

We cannot afford to have an outage during rollout of https://kubernetes.io/blog/2023/03/10/image-registry-redirect/ and we may need to be able to ship fixes during that time.

PRs that are not critical bug fixes will have to wait.

PRs that do not affect archeio (the prod registry serving code) can be accepted, e.g. improving the performance of `cmd/geranos` (which is not end user facing and will not cause an outage) or PRs which improve docs will be accepted.

BenTheElder:
We made some very small changes only to patch for golang CVE. We are still holding other changes but should lift soon.

First I'm working with Arnaud on making staging match production more closely, as we have a higher bar to ensure safety of changes now with all the additional traffic ...

BenTheElder:
The redirect is looking good, we should still proceed with caution but we have done more on that front:

- staging now shares terraform with prod
- staging now auto-deploys the terraform
- staging has more e2e tests on both GCP and AWS

I'd still like to land #226 but we can close this now.

PR #206, opened 2023-03-26
stop exporting archeio internals

BenTheElder:
exported packages that are semi-stable / reusable are already under pkg/

no code changes, just moved `cmd/archeio/app` => `cmd/archeio/internal/app`


BenTheElder:
I think this is OK despite #181, it shouldn't affect the ability to roll out bug fix releases to production.

Went ahead and filed this now because I don't want moving the package to conflict with future PRs.

BenTheElder:
Seems to be a flake, rerunning

PR #208, opened 2023-03-26
set timeout for bucket HEAD client

BenTheElder:
depends on #206 

I haven't seen evidence that this is a problem yet, but it seems obviously problematic by inspection.

There is no default timeout on http.Client. We made a similar change for the listener.

It's possible to do more granular timeouts over e.g. dialing the connection, and other tuning, but this seems like the most important and simplest change.

/hold

Not sure if this counts as a bugfix ref: #181 as currently in practice we have very good response times and haven't seen this issue yet ... but it seems bound to happen.

### DIFF
--- a/cmd/archeio/internal/app/buckets.go
+++ b/cmd/archeio/internal/app/buckets.go
@@ -19,6 +19,7 @@ package app
 import (
 	"net/http"
 	"sync"
+	"time"
 
 	"k8s.io/klog/v2"
 )
@@ -82,7 +83,7 @@ type blobChecker interface {
 // TODO: potentially replace with a caching implementation
 // should be plenty fast for now, HTTP HEAD on s3 is cheap
 type cachedBlobChecker struct {
-	http.Client
+	client *http.Client
 	blobCache
 }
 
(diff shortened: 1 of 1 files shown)


TerryHowe (approved):
/lgtm

cmd/archeio/internal/app/buckets.go
```diff
@@ -91,6 +92,15 @@ func newCachedBlobChecker() *cachedBlobChecker {
...
 			cache: make(map[string]map[string]struct{}),
 		},
+		client: newHTTPClient(),
+	}
+}
+
+func newHTTPClient() *http.Client {
+	// ensure sensible timeouts
+	// this client will be used to make HEAD calls
+	return &http.Client{
```
BenTheElder, line 102: TODO: transport and connection reuse tuning, given the large amount of calls to same-host?

BenTheElder: we do, and that will implicitly use http.DefaultTransport
http client transport can be tuned for connection idle + reuse

BenTheElder: client reuse may actually be suboptimal (locks on shared resources like cookies etc) vs just ensuring transport reuse and tuning that. 

Most advice about client re-use seems to be either related to transport reuse (which can instead be accomplished by passing a transport when constructing a client) or maybe allocation (not a primary concern here).

BenTheElder:
rebased
/hold cancel

Issue #76, opened 2022-05-31
production deployment should be gitops


BenTheElder:
We auto deploy the registry-sandbox instance, and everything for that is configured either in k8s.io terraform, or in the make rule backing `make deploy` / the cloudbuild.yml for this repo.

The production configuration currently seems to only be partially source controlled, the production (registry.k8s.io) cloud run deployment appears to still be manually submitted by @ameukam 

Now that kubernetes has migrated, we really ought to fix this https://github.com/kubernetes/kubernetes/pull/109938#event-6713039633

BenTheElder:
in addition: staging should match production as closely as possible, except for the image revision, and toggling environment variable overrides for gating new functionality

we have a limited / false sense of performance when the regions don't match because then we're talking to much more distant AWS regions https://kubernetes.slack.com/archives/CCK68P2Q2/p1659742065287909

BenTheElder:
rough outline:
- add a service account with permission to deploy this
- setup a GCB postsubmit on k8s.io repo to do terraform apply of oci-proxy-prod

BenTheElder:
we _are_ deploying from terraform in git for production, but manually.

also, we have skew in that the sandbox instance does not use the production config or terraform.

I'd like us to start by getting the sandbox instance to be the production deployment configs w/ some variable overrides passed in, and then look at automating deploying to prod once that's working.

BenTheElder:
Small change towards this in https://github.com/kubernetes/k8s.io/pull/4922

BenTheElder:
The prod deployment is totally checked in, however we're still requiring humans to roll it out and be ready to quickly revert.

That's fine, the same people we'd have approve changes have access to rollout.

Importantly: *staging* is completely automated and leads any production rollouts, which will reflect staging when we're comfortable.

Everything is in git now, other than staging automatically updating to the latest image just pushed when deploying.

PR #210, opened 2023-03-26
deploy staging by overriding image version on terraform

BenTheElder:
Part of #76 

Auto-deploying staging from the production terraform will let us build confidence that:
1. The staging environment is actually representative of production, just with newer source code
2. We can safely auto-deploy production when changes merge in the k8s.io repo (not here)

We took a shortcut on the first point in particular I'd like to fix that.

I'm less concerned about auto deploy for 2) and to be honest we should have humans on hand to observe changes when we deploy to production considering the scope of potential breakage ...

Depends on https://github.com/kubernetes/k8s.io/pull/5081

### DIFF
--- a/hack/make-rules/deploy.sh
+++ b/hack/make-rules/deploy.sh
@@ -19,43 +19,29 @@ set -o errexit -o nounset -o pipefail
 REPO_ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../.." && pwd -P)"
 cd "${REPO_ROOT}"
 
+# make sure we have a k8s.io clone for the prod terraform
+k8sio_dir="$(cd "${REPO_ROOT}"/../k8s.io && pwd -P)"
+if [[ ! -d "${k8sio_dir}" ]]; then
+    >&2 echo "Deploying requires a github.com/kubernetes/k8s.io clone at ./../k8s.io"
+    >&2 echo "FAIL"
+    exit 1
+fi
+
+# install crane and get current image digest
 TAG="${TAG:-"$(date +v%Y%m%d)-$(git describe --always --dirty)"}"
-SERVICE_BASENAME="${SERVICE_BASENAME:-k8s-infra-oci-proxy}"
+# TODO: this can't actually be overridden currently
+# the terraform always uses the default here
 IMAGE_REPO="${IMAGE_REPO:-gcr.io/k8s-staging-infra-tools/archeio}"
-PROJECT="${PROJECT:-k8s-infra-oci-proxy}"
+(cd "${REPO_ROOT}"/hack/tools && go build -o "${REPO_ROOT}"/bin/crane github.com/google/go-containerregistry/cmd/crane)
+IMAGE_DIGEST="${IMAGE_DIGEST:-$(bin/crane digest "${IMAGE_REPO}:${TAG}")}"
 
-REGIONS=(
-    asia-east1
-    asia-northeast1
-    asia-northeast2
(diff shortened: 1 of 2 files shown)


BenTheElder:
TODO: Need to update the production terraform to support a variable for archeio verbosity and then update this PR to override that.

hack/make-rules/deploy.sh
```diff
@@ -19,43 +19,31 @@ set -o errexit -o nounset -o pipefail
...
-    us-central1
-    us-east1
-    us-east4
-    us-east5
-    us-south1
-    us-west1
-    us-west2
-)
+# cd to prod terraform and apply
+cd "${k8sio_dir}"/infra/gcp/terraform/k8s-infra-oci-proxy-prod
```
ameukam, line 39: why not sure `gcloud` to get any metadata about the CR services we run in prod ? the SA could have viewer access to the prod?

BenTheElder: This deploys to staging project only

BenTheElder:
/hold
waiting for https://github.com/kubernetes/k8s.io/pull/5081

hack/make-rules/deploy.sh
```diff
@@ -19,43 +19,24 @@ set -o errexit -o nounset -o pipefail
...
-    us-east4
-    us-east5
-    us-south1
-    us-west1
-    us-west2
-)
+# cd to staging terraform and apply
+cd "${k8sio_dir}"/infra/gcp/terraform/k8s-infra-oci-proxy
+# NOTE: this must use :? expansion to ensure we will not run with unset variables
+terraform apply -auto-approve -var digest="${IMAGE_DIGEST:?}"
```
ameukam, line 41: This is not gonna be enough. you need:

```
tfswitch
terraform init 
```

before be able to apply the changes.

BenTheElder: fixed

BenTheElder:
/hold cancel
will iterate on any issues that crop up, since this only affects staging, and we're driving closer staging / prod alignment

BenTheElder:
https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-oci-proxy-push-images/1643714588077723648

BenTheElder:
Looks like working-as-intended but the GCB SA doesn't have permissions needed to access the tfstate for staging.

> Step #2 - "deploy-staging": [31m│ [31mError: Failed to get existing workspaces: querying Cloud Storage failed: googleapi: Error 403: 1017132094926@cloudbuild.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket. Permission 'storage.objects.list' denied on resource (or it may not exist)., forbidden

BenTheElder:
Otherwise correctly set up everything and terraform version + invocation looks right. We can go patch the IAM.

BenTheElder:
Running a local `make deploy TAG=v20230405-v0.2.1-6-g43f2fad` (image from this PR) deploy to ensure everything else works first with my user and that we're caught up until the IAM changes land. We shouldn't have code changes until this works anyhow.

BenTheElder:
We had some issues with reconciliation order it seems, hot-fixed by wiping some existing resources from staging manually and re-deploying. Seems to be progressing now.

BenTheElder:
https://github.com/kubernetes/k8s.io/pull/5087 follow-up

BenTheElder:
OK staging is working fine again with updated certs etc, now to figure out the GCB SA IAM

PR #212, opened 2023-03-27
add a debugging hint for GCP block

BenTheElder:
ref: https://github.com/kubernetes/registry.k8s.io/issues/211, https://github.com/kubernetes/registry.k8s.io/issues/174, https://github.com/kubernetes/registry.k8s.io/issues/138

TODO: maybe we can serve some code other than 403 from the app armor rules that *are* under Kubernetes's control (at least those should have a different reason message AFAICT)

### DIFF
--- a/docs/debugging.md
+++ b/docs/debugging.md
@@ -4,6 +4,17 @@ The registry.k8s.io is a Kubernetes container images registry that behaves gener
 
 When you are debugging issues, make sure you run these commands on the node that is attempting to run images. Things may be working fine on your laptop, but not on the Kubernetes node.
 
+<!--TODO: identify what this looks like on s3 etc.-->
+> **Note**
+>
+> If you see a [403 error][http-403] like `Your client does not have permission to get URL`,
+> this error is not specific to the Kubernetes project / registry.k8s.io and
+> you need to work with your cloud vendor / service provider to get unblocked
+> by GCP.
+>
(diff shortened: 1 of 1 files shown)


docs/debugging.md
```diff
@@ -4,6 +4,17 @@ The registry.k8s.io is a Kubernetes container images registry that behaves gener
 
 When you are debugging issues, make sure you run these commands on the node that is attempting to run images. Things may be working fine on your laptop, but not on the Kubernetes node.
 
+<!--TODO: identify what this looks like on s3 etc.-->
+> **Note**
+>
+> If you see a [403 error][http-403] like `Your client does not have permission to get URL`,
+> this error is not specific to the Kubernetes project / registry.k8s.io and
+> you need to work with your cloud vendor / service provider to get unblocked
+> by GCP.
```
TerryHowe, line 13: Some mention of "forbidden" here might help

BenTheElder: I'm trying to avoid referencing _all_ 403 errors since some 403 errors may be App Armor or S3/AWS instead, but so far not _this_ message (though, we can look into if it's reasonable to do some other 4XX from App Armor etc in the future).

BenTheElder: Unfortunately we also can't say much about this.

dims:
/approve
/lgtm

we can always iterate!

PR #216, opened 2023-04-04
improve image build times by increasing build machine size

BenTheElder:
we need to be able to build this quickly and even with a relatively cheap build the default cloudbuild machine type is single core.

custom sizes take longer to start but run faster.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -1,4 +1,6 @@
 # See https://cloud.google.com/cloud-build/docs/build-config
+options:
+  machineType: E2_HIGHCPU_8
 steps:
   - id: build-image
     name: gcr.io/cloud-builders/docker


cloudbuild.yaml
```diff
@@ -1,4 +1,6 @@
 # See https://cloud.google.com/cloud-build/docs/build-config
+options:
+  machineType: N1_HIGHCPU_8
```
TerryHowe, line 3:
```suggestion
  machineType: E2_HIGHCPU_8
```
Maybe?

BenTheElder: Yes! Old habit

BenTheElder: done

TerryHowe (approved):
/lgtm
/unhold

BenTheElder:
much better, image push step done in ~2m https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-oci-proxy-push-images/1643332044014489600

PR #218, opened 2023-04-05
log terraform command

BenTheElder:
tiny follow-up to #210 to help debugging.

Holding until we fix the terraform IAM rules, then we can merge this and test.
/hold


BenTheElder:
[refreshing jobs after test-infra update]

BenTheElder:
/hold cancel
we should be ready.
filed https://github.com/kubernetes/registry.k8s.io/pull/219 as well

hack/make-rules/deploy.sh
```diff
@@ -44,4 +45,4 @@ fi
 terraform -v
 terraform init
 # NOTE: this must use :? expansion to ensure we will not run with unset variables
-terraform apply -auto-approve -var digest="${IMAGE_DIGEST:?}"
+(set -x; terraform apply -auto-approve -var digest="${IMAGE_DIGEST:?}")
```
BenTheElder, line 48: this ensures we log exactly what we ran including the supplied `-var digest=$digest`

PR #221, opened 2023-04-06
don't set quota-project

BenTheElder:
TODO: probably revert the entire prior commit instead


BenTheElder:
also TODO: set up easier iteration on cloud build / terraform tomorrow to root cause the lingering issues.

BenTheElder:
what we can see after this is the SA doesn't "have access" to the project.
seems to be resource manager API b/c cross project. discussed with arnaud, will likely resolve with enabling the API

BenTheElder:
setting the project works fine after enabling resource manager api

PR #223, opened 2023-04-07
rewrite integration test to fake coming from different routing locations

BenTheElder:
We import crane (which is already a dependency of the main module since cmd/archeio) instead of binary invoking it and pass a transport to crane which injects a fake X-Forwarded-For header 

Since in local integration test we are NOT behind cloud run etc., there's nothing sanitizing X-Forwarded-For, and we can control the end of that header and pretend to be from any location (unlike live deployed e2e test instances).

We already test these code paths in unit tests, but this allows us to appear to pull from N locations with a real running binary in presubmit.

Resolves outstanding TODO about simulating coming from AWS outside of unit tests (since CI is primarily in GCP currently, and development typically neither GCP or AWS)

The e2e tests on AWS (kOps cluster using the staging instance) remain as well.


cmd/archeio/main_test.go
```diff
@@ -76,30 +88,108 @@ func TestIntegrationMain(t *testing.T) {
...
+		tc := testCases[i]
+		t.Run(tc.Name, func(t *testing.T) {
+			t.Parallel()
+			ref := testAddr + "/" + tc.Image
+			// ensure we supply fake IP info from test case
+			craneOpts := []crane.Option{crane.WithTransport(newFakeIPTransport(tc.FakeIP))}
+			// test fetching digest first
+			digest, err := crane.Digest(ref, craneOpts...)
+			if err != nil {
+				t.Errorf("Fetch digest for %q failed: %v", ref, err)
```
justinsb, line 104: Nit: Isn't this equivalent to Fatalf?

cmd/archeio/main_test.go
```diff
@@ -76,30 +88,108 @@ func TestIntegrationMain(t *testing.T) {
...
+			// it should not match GCP, AWS or any future providers
+			IP: "192.168.0.1",
+		},
+	}
 
-	// we're done, cleanup
-	if err := serverCmd.Process.Signal(os.Interrupt); err != nil {
-		t.Fatalf("failed to signal archeio: %v", err)
+	// generate testcases from test data
+	testCases := make([]integrationTestCase, 0, len(wellKnownImages)*len(interestingIPs))
```
justinsb, line 158: Nit: preallocating is probably not worth it in test code :-)

cmd/archeio/main_test.go
```diff
@@ -76,30 +88,108 @@ func TestIntegrationMain(t *testing.T) {
...
+	for _, image := range wellKnownImages {
+		for _, ip := range interestingIPs {
+			testCases = append(testCases, integrationTestCase{
+				Name:   fmt.Sprintf("IP:%s (%q),Image:%q", ip.Name, ip.IP, image.Name),
```
justinsb, line 162: (And FYI one trick I like to use is to make the name `fmt.Sprintf("%+v", testCase)`, though you have the digest here so it might be ugly)

justinsb (approved):

justinsb:
Some nits, nothing blocking

/approve
/lgtm

/hold for nits / others

cmd/archeio/main_test.go
BenTheElder: I tried to find something somewhat readable for the actual generated subtest names

cmd/archeio/main_test.go
BenTheElder: yeah, my copy-paste is showing a bit :-)

cmd/archeio/main_test.go
BenTheElder: fair, but is it worth removing? /shrug

cmd/archeio/main_test.go
BenTheElder: Go mangles them like this: `TestIntegrationMain/IP:GCP_("35.220.26.1"),Image:"pause:3.1"`

cmd/archeio/main_test.go
BenTheElder: Actually it's not! `Fatalf` is `Errorf` + `FailNow`, however all of the `Fail` calls are redundant to `Errorf` which is just `Logf` + `Fail`.

cmd/archeio/main_test.go
BenTheElder: removed

cmd/archeio/main_test.go
BenTheElder: dropped all the Fail

BenTheElder:
fixed nits

cmd/archeio/main_test.go
```diff
@@ -54,15 +60,22 @@ func TestIntegrationMain(t *testing.T) {
...
 	testAddr := "localhost:" + testPort
 	serverErrChan := make(chan error)
-	cmdContext, serverCancel := context.WithCancel(context.TODO())
-	serverCmd := exec.CommandContext(cmdContext, "archeio")
+	serverCmd := exec.Command("./archeio", "-v=9")
+	serverCmd.Dir = filepath.Join(rootDir, "bin")
 	serverCmd.Env = append(serverCmd.Env, "PORT="+testPort)
-	// serverCmd.Stderr = os.Stderr
-	defer serverCancel()
+	serverCmd.Stderr = os.Stderr
```
BenTheElder, line 66: If you run this with `go test -v ./cmd/archeio` you will get a lot of useful logs for debugging and for confirming the server behavior

cmd/archeio/main_test.go
justinsb: You say mangle, I say "automatically apply consistent normalization" :-)  Not a blocker!

cmd/archeio/main_test.go
```diff
@@ -76,30 +89,105 @@ func TestIntegrationMain(t *testing.T) {
...
+		},
+	}
+
+	interestingIPs := []struct {
+		Name string
+		IP   string
+	}{
+		{
+			Name: "GCP",
+			IP:   "35.220.26.1",
```
BenTheElder, line 140: TODO: very low priority: ensure these remain valid IPs (e.g. we could cross check with our own package for this)

unlikely to change + checking against the same code we'll use in the server is only slightly better than as-is

justinsb:
Thanks for addressing nits - lgtm

/lgtm

PR #225, opened 2023-04-07
add e2e tests in addition to cluster e2e testing

BenTheElder:
This will allow us to quickly test the staging instance.

After this PR we'll automatically test some quick and cheap e2e tests immediately following staging deploys.

See code comments for why these are different from the integration tests.

This PR also cleans up tracking crane versions in two places

TODO: 
- Add a presubmit test that invokes `make e2e-test-local` (e2e tests against a local instance, for testing the tests) https://github.com/kubernetes/test-infra/pull/29249
- Add periodics on GCP and AWS that invoke `make e2e-test` and add that to the docs https://github.com/kubernetes/test-infra/pull/29249 for GCP
- Expand tests

### DIFF
--- a/cmd/archeio/internal/e2e/e2e_test.go
+++ b/cmd/archeio/internal/e2e/e2e_test.go
@@ -100,6 +100,10 @@ func goInstall(t *testing.T, tool string) {
 	}
 }
 
+func binPath(name string) string {
+	return filepath.Join(binDir, name)
+}
+
 // common helper for executing test pull and checking output
 func testPull(t *testing.T, tc *testCase, pullCmd *exec.Cmd) {
 	out, err := pullCmd.CombinedOutput()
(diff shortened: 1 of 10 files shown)


cmd/archeio/internal/e2e/e2e_test.go
```diff
@@ -0,0 +1,131 @@
...
+	buildCmd := exec.Command("go", "install", tool)
+	buildCmd.Env = append(os.Environ(), "GOBIN="+binDir)
+	if out, err := buildCmd.CombinedOutput(); err != nil {
+		t.Errorf("Failed to get %q: %v", tool, err)
+		t.Error("Output:")
+		t.Fatal(string(out))
+	}
+}
+
+// common helper for executing test pull and checking output
```
BenTheElder, line 103: I have more WIP code to test other clients, which is why this is broken out already. 

Will follow up with that shortly, but I want to get the CI jobs and base concept in place first.

BenTheElder: first up https://github.com/kubernetes/registry.k8s.io/pull/226

PR #226, opened 2023-04-08
add containerd to e2e tests

BenTheElder:
With a little bit of hackery we can run containerd as the current, unprivileged user and pull images to it without unpacking.

TODO: 
- ~~expand to N stable releases (and maybe nightly?)~~ Added!
- add other critical clients where feasible (docker, cri-o ... must be able to get them running and pulling in unprivileged envrionments)


hack/tools/containerd-puller/main.go
```diff
@@ -0,0 +1,53 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// tiny utility to pull an image to containerd without unpacking it
+// TODO: see if we can ship an option in ctr images pull similar to ctr images import --no-unpack
```
BenTheElder, line 18: [because unpacking needs to create mounts in the snapshotter, and we just want to successfully pull the image into the containerd content store, we care about the image pulling code path, not the image unpacking codepath]

hack/tools/containerd-puller/go.mod
```diff
@@ -0,0 +1,50 @@
+module k8s.io/registry.k8s.io/hack/tools/containerd-puller
+
+go 1.19
+
+require github.com/containerd/containerd v1.7.0
+
+require (
```
BenTheElder, line 7: didn't want to pull all of this into the main module, so added another tools module ...

see below for: probably long term we can see about gettin `--no-unpack` in `ctr images pull` instead

hack/tools/ci-install-containerd.sh
```diff
@@ -0,0 +1,43 @@
...
+    # downlod containerd to bindir
+    wget -qO- \
+        "https://github.com/containerd/containerd/releases/download/v1.7.0/containerd-1.7.0-linux-amd64.tar.gz" \
+    | tar -C "${REPO_ROOT}/bin" -zxvf - --strip-components=1
+fi
+
+# install pull util
+(cd "${REPO_ROOT}/hack/tools/containerd-puller" && go build -o "${REPO_ROOT}/bin/containerd-puller" .)
+
+# create a config that will let the current user run rootlessly
```
BenTheElder, line 38: Todo: more accurate comment here.
We're making everything the current user which is fine because we won't actually do much. If we didn't set the uid/gid it will default to chmod to root for security reasons

dims: here it is - https://github.com/containerd/containerd/pull/8368

BenTheElder: updated this comment

hack/tools/containerd-puller/go.mod
BenTheElder: removed in favor of `ctr content fetch` TIL!

BenTheElder:
Still very fast, and we're checking v1.6.20 and v1.7.0 now (both stable releases)

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/registry.k8s.io/226/pull-registry-e2e/1644594644308201472

hack/make-rules/test.sh
```diff
@@ -38,8 +38,9 @@ go_test_opts=(
 if [[ "${MODE}" = 'unit' ]]; then
   go_test_opts+=('-tags=nointegration,noe2e')
 elif [[ "${MODE}" = 'integration' ]]; then
-  go_test_opts+=('-run' '^TestIntegration')
+  go_test_opts+=('-run' '^TestIntegration' '-tags=noe2e')
 else
+  go_test_opts+=('-tags=noe2e')
```
BenTheElder, line 43: this was a bug in #32

cmd/archeio/internal/e2e/e2e_containerd_linux_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"os"
+	"os/exec"
+	"path/filepath"
+	"testing"
+	"time"
+)
+
+func TestE2EContainerdPull(t *testing.T) {
+	t.Parallel()
+	containerdVersions := []string{"1.6.20", "1.7.0"}
```
BenTheElder, line 32: TODO: keeping these up to date ...

With crane we're floating at the latest tagged release

We can maybe figure out something reasonable to detect the most recent patch release on the current and N-1 minor versions ...

BenTheElder:
This should be a solid starting point.

upodroid (approved):
/lgtm

BenTheElder:
Interesting, this fails in post-submit because wget is not available in that image. Easy enough to fix.

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-registry-push-images/1648111849591279616

BenTheElder:
filed https://github.com/kubernetes/registry.k8s.io/pull/230 for wget => curl

PR #230, opened 2023-04-18
switch to curl for fetching binaries, rework fetching binaries

BenTheElder:
- rename containerd setup script to better reflect what it does
- consistently switch to curl for containerd and shellcheck installs (which we have more widely in our CI images and scripts)
- fold installing shellcheck into main shellcheck script and drop dockerized version in favor of host binaries aligning with other scripts

resolves https://github.com/kubernetes/registry.k8s.io/pull/226#issuecomment-1512271462 (we don't have wget in the infra-tools image, we do however have curl: https://github.com/kubernetes/k8s.io/tree/main/images/k8s-infra#contents)


BenTheElder:
oh hmm, well have to update the CI job, I guess the easiest thing to do here is leave a stub script until this merges, then drop calling the script and clean it up

BenTheElder:
OK, that'll do for now https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/registry.k8s.io/230/pull-registry-shellcheck/1648130127550222336 ✅

upodroid (approved):
/lgtm

BenTheElder:
For some reason the postsubmit didn't trigger despite being configured for all merges ...

BenTheElder:
Working on subsequent merges https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-registry-push-images/1648298079301931008

PR #231, opened 2023-04-18
simplify blob cache and improve performance*

BenTheElder:
\* will use more memory. should have better concurrency and less time spent locking.

The old implementation was written pretty quickly and overly "cleverly", attempting to dedupe the common base URL portion of the string.

We use so little memory right now that this simply isn't a concern, premature optimization :-)

However, we are almost certainly experiencing excessive contention over the blob cache RWMutex. `sync.Map` is more appropriate for append-only-cache type workloads.

Also: while we do want to configure the http client timeout, we don't want to share internal locking structures like the cookiejar, only the transport. A client that simply does not configure a transport gets the shared http.DefaultTransport anyhow.

### DIFF
--- a/cmd/archeio/internal/app/buckets.go
+++ b/cmd/archeio/internal/app/buckets.go
@@ -75,7 +75,7 @@ func awsRegionToHostURL(region, defaultURL string) string {
 type blobChecker interface {
 	// BlobExists should check that blobURL exists
 	// bucket and layerHash may be used for caching purposes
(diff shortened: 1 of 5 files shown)


BenTheElder:
NOTE: I don't have benchmarks here. Real world performance is going to be a bit different anyhow, though the current performance is already acceptable.

TerryHowe (commented):
/lgtm

TerryHowe (approved):

PR #233, opened 2023-04-25
wait for containerd to be ready to avoid CI flakes

BenTheElder:
See failures like https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/registry-sandbox-e2e-gcp/1650849058304036864

Simply due to containerd socket is not ready yet when we start pulling.

### DIFF
--- a/cmd/archeio/internal/e2e/e2e_containerd_linux_test.go
+++ b/cmd/archeio/internal/e2e/e2e_containerd_linux_test.go
@@ -94,6 +94,20 @@ func testE2EContainerdPull(t *testing.T, containerdVersion string) {
 		}
 	})
 
+	// wait for containerd to be ready
+	containerdReady := false
+	for i := 0; i < 5; i++ {
+		// nolint:gosec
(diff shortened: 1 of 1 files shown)


cmd/archeio/internal/e2e/e2e_containerd_linux_test.go
```diff
@@ -94,6 +94,19 @@ func testE2EContainerdPull(t *testing.T, containerdVersion string) {
 		}
 	})
 
+	// wait for containerd to be ready
+	containerdReady := false
+	for i := 0; i < 5; i++ {
+		if err := exec.Command("ctr", "--address="+socketAddress, "version").Run(); err == nil {
```
BenTheElder, line 100: `ctr version` prints both the client and server version like `kubectl version`, it will hang if the server is not available.

BenTheElder:
Huh, this is passing locally ... investigating

BenTheElder:
Found it, ctr in path vs the copy we just downloaded.

CI is green now https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/registry.k8s.io/233/pull-registry-e2e/1650943837092188160

upodroid (approved):
/lgtm

PR #248, opened 2023-05-13
geranos: skip already uploaded images

BenTheElder:
1. Upload the manifests **after** uploading the layers, to record that we've already uploaded a given image digest to the bucket (with an added backup of the manifest contents instead of an empty key).
2. Skip re-uploading layers when we already have the manifest uploaded.

DOES NOT serve clients from the mirrored manifests.

Should produce a pretty large speedup on the second run. We'll be able to skip making any calls for these images except 1) bulk listing references in the registry 2) one S3 call per manifest to confirm it's already uploaded.

### DIFF
--- a/cmd/geranos/s3uploader.go
+++ b/cmd/geranos/s3uploader.go
@@ -17,9 +17,15 @@ limitations under the License.
 package main
 
 import (
+	"bytes"
 	"encoding/base64"
 	"encoding/hex"
+	"errors"
+	"io"
+	"strings"
 
+	"github.com/google/go-containerregistry/pkg/crane"
+	"github.com/google/go-containerregistry/pkg/name"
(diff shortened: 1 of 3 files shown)


cmd/geranos/s3uploader.go
```diff
@@ -63,7 +69,65 @@ func newS3Uploader(dryRun bool) (*s3Uploader, error) {
...
+	for _, layer := range layers {
+		if err := s.CopyToS3(bucket, layer); err != nil {
+			return err
+		}
+	}
+	m, err := manifestBlobFromRef(ref, opts...)
+	if err != nil {
+		return err
+	}
+	return s.CopyToS3(bucket, m)
```
BenTheElder, line 82: this isn't right, this will upload to the same key prefix as layers.

PR #249, opened 2023-05-16
nit: shared log message

BenTheElder:
just log the key, it has enough context as to layer vs manifest

tiny follow-up for #248 

cc @ameukam


cmd/geranos/s3uploader.go
```diff
@@ -180,7 +180,7 @@ func (s *s3Uploader) copyToS3(bucket, key string, layer imageBlob) error {
 		uploadInput.ChecksumSHA256 = aws.String(base64.StdEncoding.EncodeToString(b))
 	}
 	// skip actually uploading if this is a dry-run, otherwise finally upload
-	klog.Infof("Uploading layer: %s", key)
+	klog.Infof("Uploading: %s", key)
```
BenTheElder, line 183: currently we log like:
`I0516 05:23:06.823144    2572 s3uploader.go:183] Uploading layer: /geranos/uploaded-images/sha256:bcf9771c0b505e68c65440474179592ffdfa98790eb54ffbf129969c5e429990`

PR #250, opened 2023-05-16
geranos cleanup

BenTheElder:
- parse klog flags
- normalize key prefixes
- parallelize skipping images


BenTheElder:
that did the trick :-)

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ar-to-s3-sync/1658530066226745344

PR #254, opened 2023-06-13
Add instructions for service providers

DKbyo:
Add instructions of what to do if your customers are getting a 403 error when connecting from your ASNs


DKbyo:
I still need some confirmations...

PR #260, opened 2023-09-21
Submit a Simplified Chinese translation file for README

Zhoneym:
Submit a Simplified Chinese translation file for README

### DIFF
--- a/README_CN.md
+++ b/README_CN.md
@@ -1,4 +1,4 @@
-# registry.k8s.io
+# registry.k8s.io 中文说明
 
 本项目实现了 registry.k8s.io 的后端，这是 Kubernetes 的容器镜像注册表。
 
--- a/README_CN.md
+++ b/README_CN.md
@@ -12,7 +12,7 @@
 
 ## 稳定性
 
-registry.k8s.io 已经GA，我们要求所有用户尽快从 k8s.gcr.io 迁移过来。
+registry.k8s.io 已经 GA，我们要求所有用户尽快从 k8s.gcr.io 迁移过来。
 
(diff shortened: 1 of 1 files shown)


BenTheElder:
Hmm, without an active program to manage this, it seems it will just fall out of sync with our existing docs.

Reaching out to SIG docs who do maintain translations for kubernetes/website.

sftim:
/hold

We don't have reviewers and approvers who can review and maintain this document; nor do we plan to.

OK to unhold if we do have a team of SIG Release approvers who can maintain Chinese and would like this README to exist with localized content.

reylejano:
I agree with @sftim that SIG Docs does not have reviewers and approvers to maintain this and is outside the scope of the zn localization team 

@Zhoneym , you can contact the SIG Docs zh localization team on the [#documentation-docs-zh on Kubernetes Slack](https://kubernetes.slack.com/archives/CE3LNFYJ1)

I'm ok to close this PR unless SIG Release would like the README localized and has reviewers/approvers to maintain

mrbobbytables:
One other potential thought,  would it be worth adding a page somewhere on k8s.io about registry.k8s.io,  the stability guarantees and basic troubleshooting? That would keep it in a place where localization drift could be tracked.

sftim:
> One other potential thought, would it be worth adding a page somewhere on k8s.io about registry.k8s.io, the stability guarantees and basic troubleshooting? That would keep it in a place where localization drift could be tracked.

Yes, that _is_ a good idea. The (English) README can link there, too.

Might end up being two pages as troubleshooting docs are separated out.

BenTheElder:
Let's plan to do that instead. Thank you all!

PR #267, opened 2023-10-17
switch to go1.21.3

BenTheElder:
Latest go. We shouldn't be affected by any of the changes in theory (notably we don't serve HTTP directly so those CVEs have not really been relevant, still good to stay up to date).


upodroid (approved):
LGTM

PR #272, opened 2024-01-20
Update IP addresses data

ameukam:
This is done to include CIDRs for newest AWS and GCP regions where registry.k8s.io will be deployed.


ameukam:
cc @BenTheElder

cmd/archeio/internal/app/buckets_test.go
```diff
@@ -27,9 +27,9 @@ func TestRegionToAWSRegionToHostURL(t *testing.T) {
 	// ensure known regions return a configured bucket
 	regions := []string{}
 	for _, ipInfo := range cloudcidrs.AllIPInfos() {
-		// AWS regions, excluding "GLOBAL" meta region and govcloud
+		// AWS regions, excluding "GLOBAL" meta region, AWS US Gov Cloud and European Soveign Cloud
```
BenTheElder, line 30: FWIW: The name is AWS GovCloud (no spaces) and they usually drop the `(US)` from the end but if it is included its `AWS GovCloud (US)` /shrug

BenTheElder (commented):
/lgtm
/approve

PR #283, opened 2024-06-19
chore: Update go version in go.mod

Okabe-Junya:
## Change List

- Update go version in go.mod

## Background

ref.  https://github.com/kubernetes/k8s.io/pull/6908#issuecomment-2179035148

I want to update the Go version of k/k8s.io. 
However, it depends on the `gcr.io/k8s-staging-infra-tools/k8s-infra` image. Furthermore, since the `gcr.io/k8s-staging-infra-tools/k8s-infra` image depends on "archeio", I want to build archeio with Go 1.22.4 (the latest stable version). To do this, I will update the go.mod of this repository.


Okabe-Junya:
I don’t know where the golang image in [`config/jobs/kubernetes/sig-k8s-infra/registry.k8s.io/sig-k8s-infra-registry-presubmits.yaml` in k/test-infra](https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/sig-k8s-infra/registry.k8s.io/sig-k8s-infra-registry-presubmits.yaml) is referenced to. (Docker Hub?)

It seems to be running with Go 1.21.3, so I want to update this (and this is causing the error).

Okabe-Junya:
/assign MadhavJivrajani 
(as the assignee of https://github.com/kubernetes/k8s.io/pull/6908)

Okabe-Junya:
FYI: https://kubernetes.slack.com/archives/CCK68P2Q2/p1718816152584299

BenTheElder:
So FYI: This go version isn't the version used to build, it's the go compatibility version, you should not update go.mod go version becaue you want to compile with a new go version, you update it when you want to use new go features.

BenTheElder:
Changing this requires more changes, because it requires tools like the linter to be updated as well.

I'm working on a similar PR, but also this PR isn't needed to update the k8s.io go versions or go.mod

Okabe-Junya:
Thank you very much!

It seems that the Go version has been updated to 284. I will close this PR.

/close

PR #290, opened 2024-07-26
switch geranos/s3uploader to aws-sdk-go-v2

dims:
Pretty straight forward port from v1 to v2. Don't know how to test locally though @BenTheElder 

xref: https://github.com/kubernetes/registry.k8s.io/issues/289


BenTheElder (commented):
Just test in dry run mode which exercises everything except successfully writing files and then we can revert if the periodic fails after this.

cmd/geranos/s3uploader.go
```diff
@@ -197,15 +199,19 @@ func keyForImageRecord(imageDigest string) string {
...
 
 func (s *s3Uploader) blobExists(bucket, key string) (bool, error) {
-	_, err := s.svc.HeadObject(&s3.HeadObjectInput{
+	_, err := s.svc.HeadObject(context.TODO(), &s3.HeadObjectInput{
 		Bucket: aws.String(bucket),
 		Key:    aws.String(key),
 	})
 	if err != nil {
-		// yes, we really have to typecast to compare against an undocument string
-		// to check if the object doesn't exist vs an error making the call
```
BenTheElder, line 206: this comment is still true? "NotFound" is an undocumented error string, this code smells but it's the only way to implement this (I would ordinarily expect in a code review to be comparing to an exported error or constant)

cmd/geranos/s3uploader.go
```diff
@@ -47,29 +49,29 @@ const blobKeyPrefix = "containers/images/"
...
-		})
-	}
-	sess, err := session.NewSession(cfg...)
+	cfg, err := config.LoadDefaultConfig(context.TODO())
 	if err != nil {
 		return nil, err
 	}
+	if dryRun {
+		// Use anonymous credentials for dry run
+		cfg.Credentials = credentials.NewStaticCredentialsProvider("", "", "")
```
BenTheElder, line 65: This is probably the part i'd want to double check, we want to be sure this is anonymous and doesn't attempt to load credentials, it's not as clear in this iteration.

cmd/geranos/s3uploader.go
dims: the `if errors.As(err, &notFound) {` should work! but i saw one report of it may be not working right, so i added the previous check as fallback

cmd/geranos/s3uploader.go
dims: fixed this after dry run testing

BenTheElder (commented):
/lgtm
/approve
thanks dims!

BenTheElder:
We've already run this, but there was nothing to sync:

https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/ar-to-s3-sync/1816893877974470656

Will have to wait for an image promotion.

PR #300, opened 2025-02-06
fix staging auto deploy

### LINKED ISSUE #299: staging instance is not updating
https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/post-registry-push-images

I think there's a gap from clickops enabling Cloud Resource Manager API thrashing vs terraform since we shipped atlantis?

(To be clear: it's good that we shipped atlantis, I think that's just why this was missed before)

Need to enable it in k8s-infra-oci-proxy.

See this failure: https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-registry-push-images/1882968475526238208

It's also currently failed more recently due to failing to find the image, which I think is related to the AR migration?

https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-registry-push-images/1887273066585657344

cc @upodroid FYI

upodroid:
Fixes #299

### DIFF
--- a/hack/make-rules/deploy.sh
+++ b/hack/make-rules/deploy.sh
@@ -37,7 +37,7 @@ fi
 TAG="${TAG:-"$(date +v%Y%m%d)-$(git describe --always --dirty)"}"
 # TODO: this can't actually be overridden currently
 # the terraform always uses the default here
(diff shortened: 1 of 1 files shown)


BenTheElder (commented):
/lgtm
/approve
thank you!

PR #288, opened 2024-07-18
regular updates: preparing for next release

BenTheElder:
just version & data bumps, so everything is fresh before tagging the next release

aiming to bring production back in sync with staging soon.

PR #301, opened 2025-02-06
Update ip ranges for Q1 2025

upodroid:
It has been a while since we updated the IP ranges #288

When do we want to start parsing ipv6?


dims:
aws-ip-ranges.json is impossible to review! 🙃 

/approve
/lgtm

/hold until we need to merge!

BenTheElder:
> aws-ip-ranges.json is impossible to review! 🙃

you could diff it versus the upstream source, we could try to enforce that but it's difficult because we don't want CI to break everytime there's new data.

I primarily look to see that it looks like valid data, the actual values at worst would just make us less efficient if they're wrong (e.g. if there was a block covering the entire IP space to one vendor)

The generated code file I believe we already lint verify, but otherwise we just want to look for malicious init type behavior right?

BenTheElder (commented):
/lgtm
/approve

upodroid:
I'll cancel the hold after I verify #300 works

xmudrii (approved):
/lgtm
/approve

upodroid:
/hold cancel

autodeploy works

PR #308, opened 2025-07-14
update go, deps, cloud IP data

BenTheElder:
EDIT: this is passing everything (tested locally as well)


BenTheElder:
https://github.com/kubernetes/registry.k8s.io/blob/main/cmd/archeio/docs/testing.md#cluster-e2e-testing

PR #309, opened 2025-07-14
add containerd 2.x to e2e tests, do not cache e2e tests

BenTheElder:
We don't need to test every client and every version, we're just trying to make sure the registry isn't doing anything that breaks in the wild clients versus our read of the spec.

That said, since we _are_ testing with this client, we should really include the current major version.


BenTheElder:
> latest version looks like

yes, I picked the latest release

unless we run into bugs with it (we haven't yet ...), I've been leaving the specific patch versions, so it's clear that the code here broke versus the code there.

BenTheElder:
(the versions also show up as junit test cases / testgrid rows, so changing versions changes which testgrid rows we have)

PR #312, opened 2025-10-27
update GCP address range

translucens:
Update GCP address range to `2025-10-26T19:11:15.703738` version.

### DIFF
--- a/cmd/archeio/internal/app/buckets.go
+++ b/cmd/archeio/internal/app/buckets.go
@@ -43,7 +43,7 @@ func awsRegionToHostURL(region, defaultURL string) string {
 	case "us-east-2", "ca-central-1":
 		return "https://prod-registry-k8s-io-us-east-2.s3.dualstack.us-east-2.amazonaws.com"
 	// US West (N. California)
-	case "us-west-1":
+	case "us-west-1", "sa-west-1":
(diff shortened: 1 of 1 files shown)


cmd/archeio/internal/app/buckets.go
```diff
@@ -43,7 +43,7 @@ func awsRegionToHostURL(region, defaultURL string) string {
 	case "us-east-2", "ca-central-1":
 		return "https://prod-registry-k8s-io-us-east-2.s3.dualstack.us-east-2.amazonaws.com"
 	// US West (N. California)
-	case "us-west-1":
+	case "us-west-1", "sa-west-1":
```
ameukam: There is a cost burden on doing this which will make us hesitant to approve this.

BenTheElder: This PR was originally only updating GCP regions.

I suggested normally we update all cloud data, and this AWS region must've been added.

I assume this was a best guess as to how to map the AWS region, there are tests ensuring all regions are mapped.

BenTheElder: Northern California is a good approximate geographically closest option from the regions we currently have. As with all of these it's possible the costs or actual routing links are sub-optimal but it seems unlikely. Not configuring it doesn't reduce cost

PR #315, opened 2025-11-30
update the ips for Q4 2025 and expand to all AWS regions

upodroid:
I updated the IPs to the most recent ranges. Last update was 5 months ago.

I also tweaked the codegen so now you can run `make codegen` from the root of the repo and it will pull the latest ip ranges & generate the go file.


upodroid:
/hold 

sa-west1 needs to be routed somewhere, I see #312 but I'm working on the S3/AR expansion project so we'll have a bucket for all the regions in Dec 2025

PR #320, opened 2026-02-01
Fix intermittent TestE2EContainerdPull failures in CI

dims:
Context: https://storage.googleapis.com/k8s-triage/index.html?job=registry-sandbox-e2e

  This PR fixes flaky failures in both registry-sandbox-e2e-gcp and registry-sandbox-e2e-aws jobs.

  Commit 1: Fix containerd e2e test readiness check and add log capture

  The v2.2.0 test was failing with "Failed to wait for containerd to be ready" after ~31s.

  Changes:
  - Capture containerd stderr and include in failure messages (previously logs were lost)
  - Detect early process exit instead of blindly waiting
  - Extend timeout from ~30s to ~55s (6 iterations of quadratic backoff)

  Commit 2: Isolate NRI socket path per test instance

  Tests run containerd 1.7.29, 2.1.5, and 2.2.0 in parallel via t.Parallel(). While most resources were isolated per-test, NRI (enabled by default in containerd 2.x) uses a global
  socket at /var/run/nri/nri.sock.

  When multiple instances start simultaneously, they race for this socket causing "address already in use" errors.

  Fix: Generate containerd config at test runtime with socket_path in the per-test temp directory, matching how we handle the main containerd socket.


dims:
@ameukam needs `/approve` as well :)

upodroid (approved):

dims:
Sweet!!! field of 🟢 

https://prow.k8s.io/?job=registry-sandbox-e2e*

/pony field of green

PR #328, opened 2026-07-29
bump deps, regenerate addresses and use the new AWS CDN

upodroid:
In prod, we'll override the aws_base_url to https://cdn.registry.k8s.io


hakman:
Tests must pass, otherwise LGTM

PR #329, opened 2026-08-01
add analytics capability to archeio

upodroid:
I added analytics capability as well as correlating a full OCI operation (get manifest, pull images, etc).

The correlation feature is very important as we'll start using Cloudfront CDN, and we need to be able to link a pull operation to all the backends we hit in the CDN and fetch blobs from the nearest one, but not necessarily the one archeio hit for the existence check.

Also, we'll start sinking the image_pull log lines to BQ/Datadog that will allow us to accurately track all image pulls.

```
jsonPayload: {
    backend: "upstream"
    caller: "app/analytics.go:70"
    cloud: "GCP"
    image: "bom/bom"
    message: "image_pull"
    reference: "sha256:6f87262882f946341503f70f92660d560eca8730db0968566e377d686536c736"
    region: "europe-west4"
    timestamp: "2026-08-01T15:12:23.818610679Z"
    traceID: "d2955b969562695cc8f3e41a543dc0fc"
    type: "blob"
    userAgent: "gcrane/v0.19.1 go-containerregistry/v0.19.1"
}
```

I took the analytics business logic idea from Istio's CF image registry proxy. https://github.com/howardjohn/registry-redirector/blob/master/src/index.ts#L237


TerryHowe (approved):
/lgtm

Put hold on in case someone else wants a look, but looks good.

xmudrii (approved):
/lgtm
/approve

PR #330, opened 2026-08-01
add azure ip ranges for analytics purposes

upodroid:
Adding Azure IP ranges so we can classify traffic from Azure for #329 

Had some AI assistance with this change.


pkg/net/cloudcidrs/internal/ranges2go/prefix_helpers.go
```diff
@@ -24,8 +24,9 @@ func dedupeSortedPrefixes(s []netip.Prefix) []netip.Prefix {
 	if l <= 1 {
 		return s
 	}
-	// for 1..len(s) if previous entry does not match, keep current
-	j := 0
+	// the first entry is always kept, for 1..len(s) if the entry does not
```
upodroid, line 27: This was detected by Fable. We were dropping the first prefix of every region.

xmudrii (approved):
/lgtm

PR #335, opened 2026-08-25
add cri-o e2e tests

upodroid:
Part of https://github.com/kubernetes/registry.k8s.io/issues/333

A bug in prod broke all cri-o/podman/skopeo users because of a bug we didn't catch with containerd.


xmudrii (approved):
/lgtm
/approve

Issue #324, opened 2026-05-13
Signature manifest digest for registry.k8s.io/etcd:v3.6.11 redirects to europe-west4 and returns MANIFEST_UNKNOWN while .sig tag succeeds via us-central1


nberlee:
### Is there an existing issue for this?

- [x] I have searched the existing issues.

### Have you checked the debugging guide [docs/debugging.md](https://github.com/kubernetes/registry.k8s.io/blob/main/docs/debugging.md)?

- [x] I have followed the debugging guide.

### What did you expect to happen?

After `registry.k8s.io` successfully resolves a cosign `.sig` tag and returns a `Docker-Content-Digest`, the same manifest should be retrievable by digest through the same logical registry name.

For:

```text
registry.k8s.io/etcd:v3.6.11
```

the image digest is:

```text
sha256:fbab3d2954652f592b2653cc1b9decdbe2a633de9320735e9f364b185b6b309a
```

The corresponding cosign signature tag is:

```text
registry.k8s.io/etcd:sha256-fbab3d2954652f592b2653cc1b9decdbe2a633de9320735e9f364b185b6b309a.sig
```

That tag resolves successfully and returns:

```text
Docker-Content-Digest: sha256:79c6de132bce22a2611de76d7007105552e3c2feb3f7effe79a58836b985c5e8
```

I would expect this to also succeed:

```text
registry.k8s.io/etcd@sha256:79c6de132bce22a2611de76d7007105552e3c2feb3f7effe79a58836b985c5e8
```

Instead, the digest request is redirected to `europe-west4-docker.pkg.dev` and returns `MANIFEST_UNKNOWN`.

### Debugging Information

This was reproduced from the affected egress path, outside Talos, using `crane`.

Request context:

```text
Date: 2026-05-13T10:55:59Z / 2026-05-13T10:57:15Z
Source IP: 185.46.234.1
registry.k8s.io DNS:
  getent hosts registry.k8s.io -> 2600:1901:0:bbc4::
  dig +short registry.k8s.io -> 34.96.108.209
```

Reproduction:

```bash
set -euxo pipefail

IMG="registry.k8s.io/etcd:v3.6.11"
REPO="registry.k8s.io/etcd"

DIGEST="$(crane digest "$IMG")"
SIGTAG="sha256-${DIGEST#sha256:}.sig"
SIGDIGEST="$(crane digest "$REPO:$SIGTAG")"

echo "IMG=$IMG"
[...]

PR #321, opened 2026-04-28
Add SIGNATURE_UPSTREAM_ENDPOINT for cosign signature routing

saschagrunert:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Adds routing logic to redirect cosign signature (.sig) and attestation (.att) manifest requests to a single canonical upstream registry instead of the per-region upstream. This allows the image promoter to write signatures to one registry and have them available globally through registry.k8s.io. Controlled via the new `SIGNATURE_UPSTREAM_ENDPOINT` environment variable; when unset, behavior is unchanged.

#### Which issue(s) this PR fixes:
Ref kubernetes-sigs/promo-tools#1762

#### Special notes for your reviewer:
Related PRs:
- kubernetes-sigs/promo-tools#1829 (remove signature replication)
- https://github.com/kubernetes/k8s.io/pull/9413 (deploy SIGNATURE_UPSTREAM_ENDPOINT)
- kubernetes/test-infra#36909 (remove Prow replication job)

#### Does this PR introduce a user-facing change?
```release-note
NONE
```

PR #336, opened 2026-09-17
Route referrers requests to the signature upstream

saschagrunert:
Route OCI referrers requests (`/v2/<name>/referrers/<digest>`) to `SIGNATURE_UPSTREAM_ENDPOINT`, like the cosign `.sig` and `.att` tag requests from #321.

The image promoter writes signatures, and now also signed promotion attestations, only to the canonical registry, so their referrers only exist there. The query string is kept, so `artifactType` filters still reach Artifact Registry. Without a signature upstream configured, referrers keep going to the regional upstream.

Nothing changes for clients yet: Cloud Armor in kubernetes/k8s.io rejects referrers requests today. Referrer manifests and blobs are fetched by digest and still go to the regional backends, so a cached fallback to the canonical upstream for those follows separately (which also covers #324). Cloud Armor should only allow referrers once both are deployed, otherwise clients outside us-central1 either see empty referrer lists or can't fetch the referrers they find.

Refs kubernetes-sigs/promo-tools#1945


cmd/archeio/internal/app/handlers.go
TerryHowe, line 63: Maybe an update for the docs?

cmd/archeio/internal/app/handlers.go
```diff
@@ -188,6 +192,18 @@ func makeV2Handler(rc RegistryConfig, blobs blobChecker, knownRepositories map[s
...
 			}
+			// referrers only exist where the signatures and attestations
+			// were pushed, keep the query to preserve artifactType filters
+			if rc.SignatureUpstreamEndpoint != "" && reReferrers.MatchString(rPath) {
+				redirectURL := signatureRedirectURL(rc, rPath)
+				if r.URL.RawQuery != "" {
+					redirectURL += "?" + r.URL.RawQuery
+				}
+				klog.V(2).InfoS("redirecting referrers request to canonical upstream", "path", rPath, "redirect", redirectURL, "traceID", traceID)
+				trackPullEvent(r, traceID, "signature-upstream", ipInfo)
```
TerryHowe: It looked like trackPullEvent would do nothing here.

cmd/archeio/internal/app/handlers.go
```diff
@@ -188,6 +192,18 @@ func makeV2Handler(rc RegistryConfig, blobs blobChecker, knownRepositories map[s
 				http.Redirect(w, r, withTraceID(redirectURL, traceID), http.StatusTemporaryRedirect)
 				return
 			}
+			// referrers only exist where the signatures and attestations
+			// were pushed, keep the query to preserve artifactType filters
+			if rc.SignatureUpstreamEndpoint != "" && reReferrers.MatchString(rPath) {
```
TerryHowe: Should this block be combinedd with the above block?

TerryHowe (changes requested):

cmd/archeio/internal/app/handlers.go
saschagrunert: Updated `docs/request-handling.md` (flow list, diagram and the signature upstream description) to cover referrers requests.

cmd/archeio/internal/app/handlers.go
saschagrunert: Good catch, `reImagePath` didn't match referrers, so no event was emitted. It now matches `referrers` with the event kind `referrers`, covered in `TestParsePullEvent`.

cmd/archeio/internal/app/handlers.go
saschagrunert: Done, both are handled in one block now, only referrers keep the query.

